PR #20, opened 2020-02-21
Wrong word

DC7IA:
I believe it should be "eincheckst" instead of "einsteigst".

Because it does not matter when the train is entered (there's limited time for that anyway) but rather when one does check in to traewelling.

[Or is that intended and I'm mis-understanding how that sentence is meant?]


HerrLevin:
> [Or is that intended and I'm mis-understanding how that sentence is meant?]

No, I think it was meant in that way but your version is more clear.

PR #26, opened 2020-02-28
remove trash from repository

MrKrisKrisu:


the-matti:
I don't think that it's a good idea to remove and ignore composer.lock and package-lock.json.

Having these in version control helps to build exactly the same setup as intended.

More information why:
- https://dev.to/gajus/stop-using-package-lock-json-or-yarn-lock-3ddi
- https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

jeyemwey:
I agree with @the-matti, please revert the deletion of `composer.lock` and `package-lock.json` and I will merge the PR.

PR #38, opened 2020-03-22
Fixing Check-In duration when stopover line is a ring

### LINKED ISSUE #37: Negative trip times
**Describe the bug**
https://traewelling.de/status/129666 (which also seems to suffer from the bug described in  https://github.com/Traewelling/traewelling/issues/10) and https://traewelling.de/status/129656 are trips with a negative journey time. The bus/train also seems to arrive at the destination before it leaves at the start.

**To Reproduce**
1. Search for `Höhenstr., Potsdam`
2. Select an 603 bus to `Platz der Einheit/West, Potsdam`
3. Check-in until `Rathaus, Potsdam` (any other stop might also work)
4. Check the logged journey time as well as the arrival and departure times

Alternatively:
1. Search for `Berlin Westkreuz`
2. Select an S 42 train to `Berlin Südkreuz (S)`
3. Check-in until `Berlin-Tempelhof` (any other stop after that might also work)
4. Check the logged journey time as well as the arrival and departure times

**Expected behavior**
The logged journey time should not be negative and the bus/train should not arrive before it has started. Also, the arrival time should be correct (the departure time is already correct).

**Additional context**
The bug seems to happen when two busses (one-way and back) or two trains (two times around the circle) are squeezed into a single hafas trip. When this happens, stations can appear multiple times in one trip. It seems like Traewelling then internally chooses the first time a station appears in the trip as the destination, rather than that one the user really chose.

jeyemwey:
Fixes #37.

Parts of this PR:
- [x] Ensuring that `departure < arrival` for all cases.
- [ ] Calculating the real distances (That does not work yet since `len(Polyline-Points) != len(Stopovers)` and the polyline points don't know when the train stops there).
  FWIW: This can become another issue, since my PR already gives a benifit and fixes a problem.


HerrLevin:
Can this be merged?

jeyemwey:
> 
> 
> Can this be merged?

Yes please

HerrLevin:
I'm a bit confused why I can't merge... On my local instance there are no conflicts.

aledjones:
Maybe the Head is too old? I recently merged a jquery update into develop...

HerrLevin:
Fuck. Did I break it?

HerrLevin:
This PR will brick the API and every Checkin-Seeder. This is big. I can't fix this atm. Sorry. :x

MrKrisKrisu:
What is the current status here?

MrKrisKrisu:
This PR is a little bit broken... I created #201 so we can continue using this patch.

PR #67, opened 2020-06-15
Fixed admin-graphs with JS-Magic. ✨

HerrLevin:


jeyemwey:
So what exactly did you do?

HerrLevin:
> So what exactly did you do?

I moved the datasets from data-tags into a script-tag, since chart.js seems to have problems with single datasets in data-tags.

jeyemwey:
Does this work in prod already?

HerrLevin:
Yes it does!

Levin Baumann

PR #81, opened 2020-08-18
Upgrade from Laravel 5.8 to 6.x

MrKrisKrisu:
The current version 5.8 of laravel doesn't get security patches since February, so I think it's time to upgrade to Laravel 6 (and maybe Laravel 7 after...) which will get Bug fixes until September 3rd, 2021 (See https://laravel.com/docs/master/releases)

I followed the guide at https://laravel.com/docs/6.x/upgrade

Actually nothing major has changed where compatibility issues should come up. The only problem which occurs is that Laravel will throw errors when the name of the argument from an route is not the same as defined in routes/api.php

I fixes this in the frontend at one point, but the API Tests throws errors until now. I currently didn't dare to touch this spot. I will look at this back, later. -> So this is currently a draft pull request.


MrKrisKrisu:
It was a little bit weird that route names are written another way as defined in the routes and declared in the function names... I've corrected this. I am amazed that it has not caused any errors so far.

PR #83, opened 2020-08-21
Underline username on leaderboard on hover

MrKrisKrisu:
...so users can see this is link. Otherwise it's just text.


jeyemwey (approved):

jeyemwey:
THanks :)

PR #84, opened 2020-08-21
Revert "Underline username on leaderboard on hover"

jeyemwey:
Reverts Traewelling/traewelling#83

Build failed for some reasons? Maybe @MrKrisKrisu can take a look into that.

### DIFF
--- a/public/css/app.css
+++ b/public/css/app.css
@@ -36387,9 +36387,3 @@ p.train-status:before {
   padding-right: 3px;
 }
 
-#leaderboard #top20 a:hover,
-#leaderboard #top20k a:hover,
-#leaderboard #top20f a:hover {
-  text-decoration: underline;
-}
-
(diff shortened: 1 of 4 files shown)


MrKrisKrisu:
Thats strange. In the PullRequest I touched just SCSS and HTML. But the test throws a `Illuminate\Database\QueryException`. I don't think my PR is the reason for. 🤨

jeyemwey:
I don't think so either, but I wonder why this issue happens now, since it did not occur on the last commits?

MrKrisKrisu:
Screenshot 2020-08-21 at 13 11 30
I just tested locally, but cannot confirm the test results...

MrKrisKrisu:
The error occurred in `database/seeds/UsersTableSeeder.php:16`. The factory for users doesn't look like there is a mechanism to check if the username is already given. So maybe the factory creates this situation exceptionally. 

Maybe @HerrLevin can look at these and fix his factory? I don't have any knowledge about Laravel factories. I generate my example data another way in Seeders...

MrKrisKrisu:
I fixed the factory, which causes the exception *sometimes* in #85. So you can close this PR and merge the other one.

PR #89, opened 2020-09-08
Run Workflows on Pull requests

MrKrisKrisu:
Workflows should be run on pull requests too


MrKrisKrisu:
The current Workflow fails because no CODACY_PROJECT_TOKEN is set.
I think it would be easier to setup codacy with an Webhook?

jeyemwey:
LGTM. Can @HerrLevin work out the codacy issues? Maybe we can drop Codacy all together?

PR #101, opened 2020-09-27
[API(v1?)] Add application name to API login

### LINKED ISSUE #71: API: /auth/login: Provide a custom clientName for the API token list
**Issue:**
https://traewelling.de/settings lists all API Tokens as belonging to the "Träwelling Personal Access Client" service. When a user has logged into Traewelling via more than one API client, this makes it hard to revoke access to a specific app. It's also unclear which apps have been granted access to the user's Traewelling account.

**Suggestion:**
Add clientName and (perhaps) URL parameters to `/auth/login` and list them in the API Token overview.

MrKrisKrisu:
Added the parameter applicationName to the API Login so the user can see which application has access to their account. This is a major change, so old applications cannot use the login endpoint. So maybe we should wait merging this (or increase the API version?)

This will fix #71.

### DIFF
--- a/api-swagger-v0.yml
+++ b/api-swagger-v0.yml
@@ -37,6 +37,10 @@ paths:
                 - password
                 - email
               properties:
+                applicationName:
+                  type: string
+                  example: Very nice app
+                  description: The name of your application
                 email:
                   type: string
                   example: gertrud@traewelling.de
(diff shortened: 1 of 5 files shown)


MrKrisKrisu:
Because @HerrLevin has another plans with the API I will close this issue. It should not be merged.

PR #104, opened 2020-09-27
Migrate to Laravel v7

MrKrisKrisu:


MrKrisKrisu:
Used the wrong base for this branch. Please use #105 instead.

PR #106, opened 2020-09-27
Catch JS error when user is logged out

MrKrisKrisu:


HerrLevin (approved):
I'll merge this without testing it locally. I might go to hell for this.

PR #108, opened 2020-09-27
Migrate to Laravel v8

MrKrisKrisu:


tests/Feature/ApiAuthTest.php
```diff
@@ -2,7 +2,6 @@
 
 namespace Tests\Feature;
 
-use Illuminate\Foundation\Testing\DatabaseMigrations;
```
HerrLevin, line 5: Is this not required?

MrKrisKrisu: Nope, it was an unused import.
The class just need/use RefreshDatabase.

HerrLevin (approved):

PR #110, opened 2020-09-28
Improve database layout, changed datatypes, added foreign keys

MrKrisKrisu:
- [x] Remove "Non-working references" in old migrations
- [x] Add Foreign keys to existing database layout
- [ ] ~~Table ``events``: Column ``trainstation`` should save the primary ID from table ``train_stations`` instead of IBNR~~
- [ ] ~~Table ``hafas_trips ``: Column ``origin`` and ``destination`` should save the primary ID from table ``train_stations`` instead of IBNR~~
- [ ] ~~Table ``hafas_trips ``: Column ``stopovers``: WTF is this? Why not using another table?~~
- [ ] ~~Table ``hafas_trips ``: Column ``polyline`` should save the primary ID from table ``poly_lines`` instead of hash~~
- [ ] ~~Table ``poly_lines ``: Column ``polyline`` should be splitted up in another tables... But this is maybe such a big thing to it later...~~
- [ ] ~~Table ``train_checkins ``: Column ``trip_id`` should save the primary ID from table ``hafas_trips`` instead of long trip ID version~~
- [ ] ~~Table ``train_checkins ``: Column ``origin`` and ``destination`` should save the primary ID from table ``train_stations`` instead of IBNR~~
- [ ] ~~Table ``train_stations ``: Column ``ibnr`` is obviouly the correct primary key, so it should be changed to ID~~
- [ ] ~~Table ``users ``: Columns ``train_distance``, ``train_duration`` and ``points`` should not saved commulated in this table. I suggest save it per checkin in another table. But this is a big thing, so maybe do it later.~~


MrKrisKrisu:
My ToDo list above is very big, so I decided to not do all of these steps in this PR. I've made some issues for this on my own Fork (https://github.com/MrKrisKrisu/traewelling/issues), so I will do there in small steps soon.

This PR is now ready for review.

MrKrisKrisu:
Hmm, the tests now fail, because SQLite doesn't support altering columns these way. But these command works without problems with MySQL and MariaDB. 
Anyone an idea how to fix this?

HerrLevin (approved):

PR #127, opened 2020-10-04
[APIv0] Add expires_at to login & signup endpoint

### LINKED ISSUE #126: API: /auth: Return login token expiration timestamp
According to the user settings page, a bearer token obtained by logging into the API expires one year after it has been obtained. The expiration timestamp or duration is not available via the API itself, so warning users that their login token is about to expire is only possible by hardcoding one year from login into an API client.

`/auth/signup` and `/auth/login` should return the expiration timestamp in addition to the bearer token.

Alternatively, the swagger docs should state that bearer tokens expire after one year (and this time shouldn't be changed without bumping the API version)

MrKrisKrisu:
closes #126

### DIFF
--- a/app/Http/Controllers/API/AuthController.php
+++ b/app/Http/Controllers/API/AuthController.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers\API;
 
 use Illuminate\Http\Request;
 use App\Http\Controllers\API\ResponseController as ResponseController;
+use Illuminate\Support\Facades\Hash;
 use Illuminate\Support\Facades\Validator;
 use Illuminate\Support\Facades\Auth;
 use App\Models\User;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/API/AuthController.php
```diff
@@ -4,62 +4,66 @@
...
         $validator = Validator::make($request->all(), [
-            'username' => 'required|string|',
-            'name' => 'required|string|',
-            'email' => 'required|string|email|unique:users',
-            'password' => 'required',
+            'username'         => 'required|string|unique:users',
+            'name'             => 'required|string|',
+            'email'            => 'required|string|email|unique:users',
+            'password'         => 'required',
             'confirm_password' => 'required|same:password'
```
MrKrisKrisu, line 21: @HerrLevin Can you explain, why it is necessary to confirm the password via API? I understand such a procedure in the frontend, but in the API?

HerrLevin: It's stupid, you're right.

HerrLevin (approved):

PR #148, opened 2020-10-18
Collision Check fixed

HerrLevin:
Fixed the collision checks and added testing for this method

### DIFF
--- a/app/Http/Controllers/TransportController.php
+++ b/app/Http/Controllers/TransportController.php
@@ -636,8 +636,12 @@ class TransportController extends Controller
      * @see https://stackoverflow.com/questions/53697172/laravel-eloquent-query-to-check-overlapping-start-and-end-datetime-fields/53697498
      */
     private static function getOverlappingCheckIns(User $user, Carbon $start, Carbon $end): Collection {
-        $start = $start->clone()->addMinutes(2);
-        $end   = $end->clone()->subMinutes(2);
+        $start = $start->addMinutes(2);
+        $end   = $end->subMinutes(2);
+        dump($start->format("Y-m-d H:i:s"));
+        dump($end->format("Y-m-d H:i:s"));
+
+
 
         $user->loadMissing(['statuses', 'statuses.trainCheckin']);
 
--- a/app/Http/Controllers/TransportController.php
+++ b/app/Http/Controllers/TransportController.php
@@ -633,17 +633,38 @@ class TransportController extends Controller
      * @param Carbon $start
      * @param Carbon $end
      * @return Collection
+     * @see https://stackoverflow.com/questions/53697172/laravel-eloquent-query-to-check-overlapping-start-and-end-datetime-fields/53697498
      */
     private static function getOverlappingCheckIns(User $user, Carbon $start, Carbon $end): Collection {
-        $start = $start->clone()->subMinutes(2);
-        $end   = $end->clone()->addMinutes(2);
+        $start = $start->clone()->addMinutes(2);
+        $end   = $end->clone()->subMinutes(2);
 
         $user->loadMissing(['statuses', 'statuses.trainCheckin']);
 
         return $user->statuses->map(function($status) {
             return $status->trainCheckin;
         })->filter(function($trainCheckIn) use ($start, $end) {
-            return $trainCheckIn->arrival->isBetween($start, $end) || $trainCheckIn->departure->isBetween($start, $end);
+            dump('XXXXXXXXXXXXXXXXXXXX');
+            dump('TD: ' . $trainCheckIn->departure . ' TA: ' . $trainCheckIn->arrival);
+            dump('D: ' . $start . ' A: ' . $end);
+            dump('T1:' . ($trainCheckIn->arrival > $start && $trainCheckIn->departure < $end));
+            dump('T2:' . ($trainCheckIn->arrival > $end && $trainCheckIn->departure < $start));
+            dump('T3:' . ($trainCheckIn->departure > $start && $trainCheckIn->arrival < $start));
+            return ($trainCheckIn->arrival > $start && $trainCheckIn->departure < $end)
+                || ($trainCheckIn->arrival > $end && $trainCheckIn->departure < $start)
+                || ($trainCheckIn->departure > $start && $trainCheckIn->arrival < $start);
+            //return $trainCheckIn->arrival->isBetween($start, $end) || $trainCheckIn->departure->isBetween($start, $end);
         });
     }
 }
+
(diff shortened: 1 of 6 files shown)


HerrLevin:
@jeyemwey we should probably not use Essen Hbf since it's causing problems.

jeyemwey:
> @jeyemwey we should probably not use Essen Hbf since it's causing problems.

Oof yes, Essen Hbf does not receive the usual traffic due to the fire last month.

I will add my remarks tomorrow :)

app/Http/Controllers/TransportController.php
```diff
@@ -633,17 +633,22 @@ public static function usageByDay(Carbon $date) {
...
+        $end   = $end->subMinutes(2);
+
         return $user->statuses->map(function($status) {
             return $status->trainCheckin;
         })->filter(function($trainCheckIn) use ($start, $end) {
-            return $trainCheckIn->arrival->isBetween($start, $end) || $trainCheckIn->departure->isBetween($start, $end);
+            return ($trainCheckIn->arrival > $start && $trainCheckIn->departure < $end)
+                || ($trainCheckIn->arrival > $end && $trainCheckIn->departure < $start)
+                || ($trainCheckIn->departure > $start && $trainCheckIn->arrival < $start);
         });
```
jeyemwey: Ist das ein Rollback von Kris's Änderungen?

tests/Feature/CheckinTest.php
```diff
@@ -191,10 +192,114 @@ public function testCheckin() {
...
+            $trips[0]->destination,
+            '',
+            $user,
+            0,
+            0,
+            0,
+            0
+        );
+
+        for ($i = 1; $i < 5; $i++) {
```
jeyemwey: Wieso 1-5? Vielleicht auf mehrere Arrays aufteilen, die dann von vorne bis hinten geprüft werden?

tests/Feature/CheckinTest.php
```diff
@@ -191,10 +192,114 @@ public function testCheckin() {
...
+                    0
+                );
+                $this->fail("Expected exception for Collision Case $i not thrown");
+            } catch (CheckInCollisionException $e) {
+                $this->assertEquals($trips[0]->linename, $e->getCollision()->HafasTrip->first()->linename);
+            }
+        }
+
+        //check normal checkin possibility
+        for ($i = 5; $i < 7; $i++) {
```
jeyemwey: s.o.

jeyemwey (changes requested):

app/Http/Controllers/TransportController.php
MrKrisKrisu: Yes and no. I forgot a case in my code, Levin has added it now with these. (See Case 3 in his test cases)

app/Http/Controllers/TransportController.php
```diff
@@ -633,17 +633,22 @@ public static function usageByDay(Carbon $date) {
...
 
+        //increase the tolerance for start and end of collisions
+        $start = $start->addMinutes(2);
+        $end   = $end->subMinutes(2);
+
         return $user->statuses->map(function($status) {
             return $status->trainCheckin;
         })->filter(function($trainCheckIn) use ($start, $end) {
-            return $trainCheckIn->arrival->isBetween($start, $end) || $trainCheckIn->departure->isBetween($start, $end);
+            return ($trainCheckIn->arrival > $start && $trainCheckIn->departure < $end)
```
MrKrisKrisu: I recommend using the Carbon methods ->isAfter and ->isBefore.

database/factories/HafasTripFactory.php
```diff
@@ -117,4 +113,35 @@ public function definition() {
...
+        return $this->afterCreating(function(HafasTrip $hafasTrip) {
+            $stopOvers = json_decode($hafasTrip->stopovers);
+            $startTime = $hafasTrip->departure;
+            $endTime   = $hafasTrip->arrival;
+            $cnt       = count($stopOvers);
+            $interval  = (strtotime($endTime) - strtotime($startTime)) / $cnt;
+            $add       = $interval;
+
+            for ($i = 1; $i < $cnt; $i++) {
+                $stopOvers[$i]->arrival   = date('Y-m-d H:i:s', strtotime($startTime) + $add);
```
MrKrisKrisu: $startTime is an Carbon object. So why do you put it in an strtotime to format it with date?

Just use $startTime->format('Y-m-d H:i:s').

MrKrisKrisu: (Same with $endTime)

MrKrisKrisu: Is it necessary to format these? Why dont just save the Carbon object in the $stopOvers? This should be formatted in ISO 8601 automatically on json_encode.

database/factories/HafasTripFactory.php
```diff
@@ -117,4 +113,35 @@ public function definition() {
...
+                $add                      += $interval;
+            }
+            $stopOvers[0]->arrival          = date('Y-m-d H:i:s', strtotime($startTime));
+            $stopOvers[0]->departure        = date('Y-m-d H:i:s', strtotime($startTime));
+            $stopOvers[$cnt - 1]->arrival   = date('Y-m-d H:i:s', strtotime($endTime));
+            $stopOvers[$cnt - 1]->departure = date('Y-m-d H:i:s', strtotime($endTime));
+
+
+            $hafasTrip->stopovers = json_encode($stopOvers);
+            $hafasTrip->save();
```
MrKrisKrisu: I recommend using $hafasTrip->update(['stopovers' => xxx]);

tests/Feature/CheckinTest.php
MrKrisKrisu: Or just make an foreach with $trips.

tests/Feature/CheckinTest.php
MrKrisKrisu: Or just make an foreach with $trips.

MrKrisKrisu (changes requested):

database/factories/HafasTripFactory.php
HerrLevin: Oh wait, you can do that? I thought, I should format it so there won't be any complications.

MrKrisKrisu: Yeah, I think so, because the when you do the JSON Encode it will try to parse the toString-Method of an Object.
https://github.com/briannesbitt/Carbon/blob/db2173fbb9ebb97645c17ea23b1ec0c9be6abb0d/src/Carbon/Traits/Converter.php#L115

PR #150, opened 2020-10-19
Optimize cache

HerrLevin:
Optimized the .gitignores with excluding the bootstrap/cache files since this is automatically created and should not exist in the repo. Also correctly refreshed the .gitignore-dependencies.


artisan
```diff
@@ -1,53 +0,0 @@
-#!/usr/bin/env php
```
MrKrisKrisu: Why do you remove ``artisan``? 
I think this should be part of the repo.

MrKrisKrisu (changes requested):

MrKrisKrisu (approved):

MrKrisKrisu:
The repo just got cleaner!

CleanUp

PR #155, opened 2020-11-11
Code Cleanup

HerrLevin:


app/Http/Controllers/API/NotificationController.php
```diff
@@ -3,43 +3,39 @@
...
     }
 
     /**
      * Remove the specified resource from storage.
      *
-     * @param  int $notificationID
+     * @param int $notificationID
      *
-     * @return \Illuminate\Http\Response
+     * @return JsonResponse
```
MrKrisKrisu, line 36: Can you add ``JsonResponse`` to the function, so it become Typesafe? (Just a wish from my side.)

app/Http/Controllers/SocialController.php
```diff
@@ -117,13 +126,16 @@ public function callback($provider)
      * @param $getInfo (response of Socialite->user())
      * @param $provider (String of login-provider)
      *
-     * @return user model
+     * @return RedirectResponse|Authenticatable model
```
MrKrisKrisu, line 129: ``RedirectResponse`` or ``Authenticatable``? What?

Yes, I see it in line 132 and 183, but thats weird... We should either return and Reponse or a Model.

app/Traits/Encryptable.php
```diff
@@ -11,7 +10,7 @@ trait Encryptable
      *
      * @param  $key
      *
-     * @return $value
+     * @return mixed|string $value
```
MrKrisKrisu, line 13: ``mixed`` already contains ``string``, so I would remove the ``string``part.

MrKrisKrisu:
done in #183

PR #156, opened 2020-12-02
✨Private profiles

HerrLevin:
This pull request implements a private profile wherein check ins will not be seen publicly. This does NOT implement the possibility to follow other people and see their statuses afterwards. This will come in a future pull request.

Fixes some parts of #1.


.idea/deployment.xml
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="PublishConfigData" serverName="localhost:8000" />
+  <component name="PublishConfigData" serverName="Lumen" />
```
MrKrisKrisu, line 3: Maybe exclude this file from git?

api-swagger-v0.yml
```diff
@@ -1399,6 +1403,9 @@ components:
           type: string
           format: integer
           example: "66"
+        private_profil:
```
MrKrisKrisu: Not "profile"? (with e)

app/Http/Controllers/API/UserController.php
```diff
@@ -17,7 +17,7 @@ public function show ($username)
 
     public function active($username)
     {
-        $user           = User::where('username', $username)->firstOrFail();
+        $user           = User::where('username', 'LIKE', $username)->firstOrFail();
```
MrKrisKrisu, line 20: Why do you need "LIKE" here? If you dont use space characters as % or _ you should use = so the query is faster.

app/Http/Controllers/StatusController.php
```diff
@@ -19,14 +19,27 @@
...
+        //PrivateProfile change to "also following"
+        if (Auth::check()) {
+            $authID = Auth::user()->id;
+        }
+
+        if ($status->user->id == $authID || !$status->user->private_profile) {
+            return $status;
+        }
+
+        abort(403);
```
MrKrisKrisu, line 41:
```suggestion
        //PrivateProfile change to "also following"
        if($status->user->private_profile && (!Auth::check() || Auth::user()->id != $status->user->id)) {
           abort(403);
        }

        return $status;
```

app/Http/Controllers/StatusController.php
```diff
@@ -36,44 +49,56 @@ public static function getStatus($statusId)
...
+                                     'trainCheckin.HafasTrip',
+                                     'event')
+                              ->withCount('likes')
+                              ->whereHas('trainCheckin', function($query) {
+                                  $query->where('departure', '<', date('Y-m-d H:i:s'))
+                                        ->where('arrival', '>', date('Y-m-d H:i:s'));
+                              })
+                //PrivateProfile This needs to be removed with the Follow-Request-Feature
+                              ->whereHas('user', function($query) use ($authID) {
+                                  return $query->where('private_profile', false)->orwhere('id', $authID);
```
MrKrisKrisu:
```suggestion
                                  return $query->where('private_profile', false)->orWhere('id', $authID);
```

app/Http/Controllers/StatusController.php
```diff
@@ -19,14 +19,27 @@
...
 {
-    public static function getStatus($statusId)
-    {
-        return Status::where('id', $statusId)->with('user',
-                                                    'trainCheckin',
-                                                    'trainCheckin.Origin',
-                                                    'trainCheckin.Destination',
-                                                    'trainCheckin.HafasTrip',
-                                                    'event')->withCount('likes')->firstOrFail();
+    public static function getStatus($statusId) {
```
MrKrisKrisu:
```suggestion
    public static function getStatus(int $statusId): Status {
```

database/migrations/2020_11_18_101929_add_user_private_profile_key.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table("users", function (Blueprint $table) {
+            $table->boolean('private_profile')->default(false);
```
MrKrisKrisu:
```suggestion
            $table->boolean('private_profile')->default(false)->after('home_id');
```

Please specify where the column should be created. Otherwise it will be created after created_at and updated_at.

database/migrations/2020_11_18_101929_add_user_private_profile_key.php
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::dropIfExists('private_profile');
```
MrKrisKrisu:
```suggestion
        Schema::table('users', function (Blueprint $table) {
            $table->dropColumn('private_profile');
        });
```

Your Code tries to delete the table ``private_profile`` which doesn't exist.

resources/views/profile.blade.php
```diff
@@ -9,22 +9,25 @@
...
                     <h2 class="card-title h1-responsive font-bold">
-                        <strong>{{ __('profile.statistics-for') }} {{ $user->name }}</strong> <small class="font-weight-light">{{ '@'.$user->username }}</small>
-                        @if($currentUser)
-                            @if($user->id !== $currentUser->id && Auth::check())
-                                @if(Auth::user()->follows->where('id', $user->id)->first() === null)
-                                    <a href="#" class="btn btn-sm btn-primary follow" data-userid="{{ $user->id }}" data-following="no">{{__('profile.follow')}}</a>
-                                @else
-                                    <a href="#" class="btn btn-sm btn-danger follow" data-userid="{{ $user->id }}" data-following="yes">{{__('profile.unfollow')}}</a>
+                        <strong>{{ $user->name }} @if($user->private_profile) <i class="fas fa-user-lock"></i>@endif </strong> <br/>
+                        <small class="font-weight-light">{{ '@'.$user->username }}</small>
```
MrKrisKrisu:
```suggestion
                        <small class="font-weight-light">@{{ $user->username }}</small>
```

resources/views/search.blade.php
```diff
@@ -26,7 +26,7 @@
 
                             <div class="col pl-0">
                                 <a href="{{ route('account.show', ['username' => $user->username]) }}">
-                                    <h4>{{ $user->name }} <small>{{ $user->username }}</small></h4>
+                                    <h4>{{ $user->name }}@if($user->private_profile)<i class="fas fa-user-lock"></i>@endif <small class="text-muted">{{ '@'.$user->username }}</small></h4>
```
MrKrisKrisu, line 29:
```suggestion
                                    <h4>{{ $user->name }}@if($user->private_profile)<i class="fas fa-user-lock"></i>@endif <small class="text-muted">@{{ $user->username }}</small></h4>
```

MrKrisKrisu (changes requested):

api-swagger-v0.yml
HerrLevin: Oh, yes.

app/Http/Controllers/StatusController.php
HerrLevin: This will not work since `Auth::user()` has no attribute `id` if a profile page visitor is not logged in. Thats why the variable `authID` is needed with a null initialization.

MrKrisKrisu: But thats why ```!Auth::check()``` is checked before?
If user isn't logged in (```!Auth::check()``` is true) then the second condition doesn't get checked.

resources/views/search.blade.php
HerrLevin: Will not work. This will cause blade to stop working.

app/Http/Controllers/StatusController.php
HerrLevin: This will only work in an AND-case. OR doesn't work that way. If `Auth::check` equals false, `Auth::user()->id != $status->user->id` might still return true which is why this method always will be accessed.

MrKrisKrisu: If the User isn't logged in the first condition will return ```true```, so the second condition doesn't get checked.
It the User is logged in, the first condition is ```false```and the ```Auth::user()->id``` is called.

See: 
Screen Shot 2020-12-02 at 13 18 18
Source: https://www.php.net/manual/de/language.operators.logical.php

HerrLevin: Ah. Oh. Yeah, I'm sorry. This makes sense. For some reason I thought, `!Auth::check()` equals to false, not true... smh :man_facepalming:

MrKrisKrisu: giphy

MrKrisKrisu (approved):

HerrLevin:
Since it needs to be checked that the private profile feature conforms with the privacy policy and vice versa, it cannot be merged until this is confirmed.

MrKrisKrisu:
> Since it needs to be checked that the private profile feature conforms with the privacy policy and vice versa, it cannot be merged until this is confirmed.

Can you explain why we are waiting for this? This feature gives more privacy to the users, so why should it be blocking?

PR #167, opened 2021-01-25
Test all supported PHP versions

MrKrisKrisu:
The new GitHub Action will test the code on ~~all~~ supported PHP Versions (~~PHP 7.3~~, ~~PHP 7.4~~ & PHP 8.0)


.github/workflows/phpunit.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          extensions: mbstring, bcmath, ctype, fileinfo, json, openssl, pdo, tokenizer, xml
```
HerrLevin: Shouldn't we also use the `gd` extension?

HerrLevin (commented):
I'm fine with this but I'm unsure with the `gd`-extension.

HerrLevin (approved):

PR #183, opened 2021-01-29
♻️ Type safety & Code Formatting

MrKrisKrisu:
I thought we had cleaned up all the code before. But we didn't, so I've done that now.

The first two commits I made the functions type safe by hand. The last commit is autoformatting from the IDE.


HerrLevin (approved):
bassd scho

PR #201, opened 2021-01-31
Fixing Check-In duration when stopover line is a ring

### LINKED ISSUE #37: Negative trip times

MrKrisKrisu:
Child PR from #38 
This is forked from `dev/jv/37-negative-trip-times`

fixes #37


app/Models/TrainCheckin.php
```diff
@@ -93,12 +94,15 @@ public function getMapLines() {
...
-    }
+    public function getDurationAttribute(): int {
+        try {
+            return $this->origin_stopover->departure_planned->diffInMinutes(
+                $this->destination_stopover->arrival_planned
+            );
+        } catch (Exception) {
+            //We need the try-catch to support old checkins, where no stopovers are saved.
+            return $this->arrival->diffInMinutes($this->departure);
+        }
```
jeyemwey: Also DA hättest du dir auch denken können, dass die Codacy-Götter das nicht durchlassen.

MrKrisKrisu: Sorry. xD

aledjones:
Finally! 🎉

HerrLevin:
Since the new business logic now requires a supplied departure and arrival time in the check in process, ring lines will still be broken in the API. This needs to be fixed in a future™ update b/c Kris despises the API and doesn't want to work with it.

MrKrisKrisu:
> Since the new business logic now requires a supplied departure and arrival time in the check in process, ring lines will still be broken in the API. This needs to be fixed in a future™ update b/c Kris despises the API and doesn't want to work with it.

I don't despise it, I just don't support it.

HerrLevin (approved):

PR #210, opened 2021-02-14
🩹 Fixed ring-lines in API

### LINKED ISSUE #202: Fix ring lines in API
Since the new business logic now requires a supplied departure and arrival time in the check in process, ring lines will still be broken in the API. This needs to be fixed in a future™ update b/c Kris despises the API and doesn't want to work with it.

_Originally posted by @HerrLevin in https://github.com/Traewelling/traewelling/issues/201#issuecomment-774339928_

MrKrisKrisu:
fixes #202

The new attributes are nullable so this is not a breaking change. The API is functional without these parameters but maybe inaccurate at ring-lines.

### DIFF
--- a/api-swagger-v0.yml
+++ b/api-swagger-v0.yml
@@ -584,6 +584,14 @@ paths:
                   type: boolean
                 toot:
                   type: boolean
+                departure:
+                  description: Time of departure at the start station
+                  nullable: true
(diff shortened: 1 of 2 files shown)


HerrLevin (approved):
This needs to be added to the self-generating API-documentation

HerrLevin:
This needs to be added to the self-generating API-documentation

PR #213, opened 2021-02-14
🩹 Check if account has set an e-mail address before removing sso

### LINKED ISSUE #137: Check if Account has set an E-Mail Address
Removing the SSO provider checks that the account has a password set, but does not that an email address is set.

It should be checked as a user without an email address but  with a password cannot log in... :D

MrKrisKrisu:
fixes #137

### DIFF
--- a/resources/lang/en/controller.php
+++ b/resources/lang/en/controller.php
@@ -4,7 +4,7 @@ return [
         "already-connected-error" => "This Account is already connected to another user",
         "create-error"            => "There has been an error creating your account.",
         "delete-never-connected"  => "Your user does not have a Social Login provider",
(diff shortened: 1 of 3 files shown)


resources/lang/en/controller.php
```diff
@@ -4,6 +4,7 @@
         "already-connected-error" => "This Account is already connected to another user",
         "create-error"            => "There has been an error creating your account.",
         "delete-never-connected"  => "Your user does not have a Social Login provider",
+        "delete-set-email"        => "You need to set a E-Mail Address before deleting a SSO-Provider to prevent you from locking yourself out.",
```
HerrLevin: Before you delete an SSO provider, you need to set an email address so that you don't get locked out.

HerrLevin (changes requested):

HerrLevin (approved):

PR #216, opened 2021-02-15
✨ Export journeys as ICS

### LINKED ISSUE #14: Export journeys as ICS
I love myself a tidy calendar that has all my checked-in journeys. I'm using subscribable calendars already and know how to deal with them.

Offer `.ics` files as an export option for a given timespan, comparable with the export options that exist.

Offer a tokenized link that my calendar app of choice can read periodically. Issue here is, that those files become very large. Maybe a mix of both options?

**Additional context**
* [RFC 5545 describing iCal](https://tools.ietf.org/html/rfc5545)
* [StackOverflow question asking about large ics files](https://stackoverflow.com/questions/59685510/how-to-deal-with-large-ics-files)
* Related Gitlab [issue](https://gitlab.com/HerrLevin_/trwl/issues/176) and [merge request](https://gitlab.com/HerrLevin_/trwl/-/merge_requests/180)

MrKrisKrisu:
closes #14


resources/lang/en/settings.php
```diff
@@ -1,42 +1,49 @@
...
+    "title-sessions"              => "Sessions",
+    "title-tokens"                => "API-Tokens",
+    "title-ics"                   => "Export to calendar",
+    "upload-image"                => "Upload profile avatar",
+    "client-name"                 => "Service",
+    "created"                     => "Created",
+    "updated"                     => "Updated",
+    "expires"                     => "Expires",
+    "deletetokenfor"              => "Invalidate token for:",
+    "no-tokens"                   => "There a no external applications with access to your account.",
```
HerrLevin: There are no third-party applications with access to your account.

resources/lang/en/settings.php
```diff
@@ -1,42 +1,49 @@
...
+    "no-ics-tokens"               => "You haven't created a calendar share for your journeys yet.",
+    "token"                       => "Token",
+    "revoke-token"                => "Revoke access",
+    "create-ics-token"            => "Create new token",
+    "create-ics-token-success"    => "The Calendar-Token has been created. You can embed it with the following link to every software which support ICS-Files: :link",
```
HerrLevin: `calendar token` and `ICS files`

resources/lang/en/settings.php
```diff
@@ -1,42 +1,49 @@
...
+    "revoke-ics-token-success"    => "The Calendar-Token has been revoked.",
```
HerrLevin: calendar token

app/Http/Controllers/IcsController.php
```diff
@@ -0,0 +1,114 @@
...
+    public function renderIcs(Request $request): Response {
+        $validated = $request->validate([
+                                            'user_id' => ['required', 'exists:users,id'],
+                                            'token'   => ['required', 'exists:ics_tokens,token'],
+                                            'limit'   => ['nullable', 'numeric', 'gte:1', 'lte:10000'],
+                                            'from'    => ['nullable', 'date'],
+                                            'to'      => ['nullable', 'date']
+                                        ]);
+
+        $icsToken = IcsToken::where('token', $validated['token'])->first();
```
HerrLevin, line 25: could this be done with a `firstOrFail()`?

app/Http/Controllers/IcsController.php
```diff
@@ -0,0 +1,114 @@
...
+
+class IcsController extends Controller
+{
+    public function renderIcs(Request $request): Response {
+        $validated = $request->validate([
+                                            'user_id' => ['required', 'exists:users,id'],
+                                            'token'   => ['required', 'exists:ics_tokens,token'],
+                                            'limit'   => ['nullable', 'numeric', 'gte:1', 'lte:10000'],
+                                            'from'    => ['nullable', 'date'],
+                                            'to'      => ['nullable', 'date']
```
HerrLevin: would be better if `to` can be renamed to `until` so it's definitely time-related. In the first moment I thought this has something to do with trains & destinations. 😅

HerrLevin (changes requested):

app/Http/Controllers/IcsController.php
MrKrisKrisu: This is not needed because it is already checked in the validator.

HerrLevin: So then the first comparison in Line 27 is kinda redundant, isn't it?

MrKrisKrisu: The `== null` part is redundant, yes. But the check against the user_id is needed.

HerrLevin: Okay. I'll approve this PR but please remove the `$icsToken == null ||`. You can then merge it yourself. :)

HerrLevin (approved):

Issue #7, opened 2020-02-19
Marking commuter journeys, railroad workers, private trips


jeyemwey:
**Is your feature request related to a problem? Please describe.**
* As a commuter, I'd like to track my journeys for tax reasons.
* As a railroad worker (Tf, ZuB, KIN, etc.), I'd like to seperate my work from the rest of my trips. I, too, have deducible journeys that I'd like to track for tax reasons.
* As a private traveller, I'd like to keep my holiday trips out of my tax returns.

**Describe the solution you'd like**
On check-in (and later), I'd like to flag my status as a commuter trip or as a work trip. When I export my data, I'd like to have a filter.

**Describe alternatives you've considered**

--

**Additional context**

TRWL beta had this feature, but it was removed before the release. The feature did not seperate between commuters and rail workers, but rather just splitted between business and private trips. See `db9362cf41d08211e4ecb439b2e94dccc2b1d6b4` which reverts the frontend code.

[Gitlab issue regarding the feature removal and how we could split trips](https://gitlab.com/HerrLevin_/trwl/issues/118)

PR #243, opened 2021-04-13
📦️ Bootstrap5

HerrLevin:
This is an upgrade to Bootstrap5. Please run this as a test environment and test all functions of the Webinterface.

Also removes the workaround of #28 and fixes it! (With a fancy animation :bell: )

Fixes some not fixed things from #220

PR #248, opened 2021-04-28
✨ Business CheckIn

HerrLevin:
Finally you can categorize your checkin by business, private and commute.


HerrLevin:
⚠️ This can be merged when #243 is merged. 

Closes #7

app/Http/Controllers/FrontendStatusController.php
```diff
@@ -64,14 +64,14 @@ public function DeleteStatus(Request $request): JsonResponse|RedirectResponse {
 
     public function EditStatus(Request $request): JsonResponse|RedirectResponse {
         $this->validate($request, [
-            'body'          => 'max:280',
-            'businessCheck' => 'max:1',
+            'body'           => 'max:280',
+            'business_check' => 'digits_between:0,2'
```
MrKrisKrisu:
```suggestion
            'body'           => ['required', 'max:280'],
            'business_check' => ['required', 'digits_between:0,2'],
```
Would be nice if we define if it is required or nullable

app/Http/Controllers/StatusController.php
```diff
@@ -299,7 +301,8 @@ public static function ExportStatuses($startDate,
                                       'Kilometer',
                                       'Punkte',
                                       'Status',
-                                      'Zwischenhalte'
+                                      'Zwischenhalte',
+                                      'Fahrttyp'
```
MrKrisKrisu, line 305: Translation?

database/migrations/2021_02_23_185026_change_avatar_to_be_null_on_users.php
```diff
@@ -1,4 +1,4 @@
-<?php
+<?php1
```
MrKrisKrisu:
```suggestion
<?php
```

image

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
```diff
@@ -0,0 +1,41 @@
...
+
+        DB::table('statuses')
+            ->whereNull('business')
+            ->update(['business' => '0']);
+        DB::table('statuses')
+            ->where('business', 'true')
+            ->update(['business' => '1']);
+        DB::table('statuses')
+            ->where('business', 'false')
+            ->update(['business' => '0']);
```
MrKrisKrisu:
```suggestion
```
We don't have "true" or "false" in our database. There are only the numbers 0 and 1, so in my opinion this migration is unnecessary.

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
```diff
@@ -0,0 +1,41 @@
...
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+class ChangeStatusesToAcceptBusinessCheckinAsInteger extends Migration
+{
+    public function up() {
+        Schema::table('statuses', function(Blueprint $table) {
+            $table->integer('business')
+                  ->default(0)
+                  ->change();
```
MrKrisKrisu:
```suggestion
            $table->unsignedTinyInteger('business')
                  ->default(0)
                  ->change();
```
Storage space costs money! xD

An unsigned tiny integer is enough. The range goes from 0 to 255.

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
```diff
@@ -0,0 +1,41 @@
...
+            $table->boolean('business')
+                ->change();
+        });
+
+        DB::table('statuses')
+            ->where('business', '>=', '1')
+            ->update(['business' => true]);
+        DB::table('statuses')
+            ->where('business', '0')
+            ->update(['business' => false]);
```
MrKrisKrisu:
```suggestion
```
See the last comment for this...

resources/js/components/business-check-in.js
```diff
@@ -0,0 +1,82 @@
...
+$(businessBriefcase).on("click", function (){
+    setIconsForCheckIn(1)
+})
+
+$(businessBuilding).on("click", function (){
+    setIconsForCheckIn(2)
+})
+
+$(document).on("click", ".edit", function (event) {
+    console.log("edit");
```
MrKrisKrisu:
```suggestion
```

image

resources/sass/_variables.scss
```diff
@@ -7,6 +7,9 @@ $font-size-base: 0.9rem;
 $line-height-base: 1.6;
 
 // Colors
+$white: #fff;
+$black: #000;
```
MrKrisKrisu: Why do you define black and white when you don't use them yourselves? :D

resources/sass/components/custom-buttons.scss
```diff
@@ -0,0 +1,37 @@
+.btn-outline-twitter {
+  color: #1da1f2;
+  border-color: #1da1f2;
+}
+
+.btn-twitter {
+  color: #fff;
```
MrKrisKrisu: e.g.

resources/sass/components/custom-buttons.scss
```diff
@@ -0,0 +1,37 @@
...
+  background-color: #1da1f2;
+}
+
+.btn-check:active + .btn-outline-twitter,
+.btn-check:checked + .btn-outline-twitter,
+.btn-outline-twitter.active,
+.btn-outline-twitter:active,
+.show > .btn-outline-twitter.dropdown-toggle {
+  color: #fff;
```
MrKrisKrisu: e.g.

resources/views/includes/edit-modal.blade.php
```diff
@@ -3,18 +3,37 @@
...
+                               aria-expanded="false"
+                        >
+                            <i class="fa fa-user" id="business-user"></i>
+                            <i class="fa fa-briefcase d-none" id="business-briefcase"></i>
+                            <i class="fa fa-building d-none" id="business-building"></i>
+                        </label>
+                        <ul class="dropdown-menu" aria-labelledby="dropdownMenuButton">
+                            <li class="dropdown-item" id="business-li-user"><i class="fa fa-user"></i> Normaler User Check-in</li>
+                            <li class="dropdown-item" id="business-li-briefcase"><i class="fa fa-briefcase"></i> Geschäftlicher Check-in (für die Arbeit)</li>
+                            <li class="dropdown-item" id="business-li-building"><i class="fa fa-building"></i> Arbeitsweg Check-in (für eine Strecke zum Arbeitsplatz)</li>
```
MrKrisKrisu: Translations?

resources/views/layouts/app.blade.php
```diff
@@ -107,25 +112,23 @@
...
+                                       data-mdb-toggle="modal"
+                                       data-mdb-target="#notifications-board">
                                         <span class="notifications-bell far fa-bell"></span>
                                     </a>
                                 </li>
                                 <li class="nav-item dropdown">
-                                    <a id="navbarDropdown" class="nav-link dropdown-toggle mdb-select"
-                                       href="javascript:void(0)"
-                                       role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                                    <a id="navbarDropdown" href="#" class="nav-link dropdown-toggle mdb-select"
```
MrKrisKrisu: Why do you changed this back to `href="#"?`

`javascript:void(0)` would be result in a better user experience.

MrKrisKrisu (changes requested):

app/Http/Controllers/StatusController.php
HerrLevin: So here's one big problem: The csv never has been translated and I'm not sure if it's a good idea to change that __now__, because this would cause breaking changes for some users? Idk.

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
HerrLevin: It's actually defined as boolean in the first database migration, so I'm unsure about that.

resources/sass/_variables.scss
HerrLevin: Otherwise this caused a bug in the SASS-compiler. This is a remnant of #243. :D

resources/views/layouts/app.blade.php
HerrLevin: This is a remnant of #243. Won't work otherwise, afaik/iirc.

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
HerrLevin: We can't remove this, since the business-column is defined as a boolean, so we need to store everything greater than 1 as a 1. It's debatable whether we need to define it as true and false, b/c sqlite and afaik MariaDB define it as normal integers.

app/Http/Controllers/StatusController.php
MrKrisKrisu: Okay, but then it would be nice if we create an issue for this.

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
MrKrisKrisu: A boolean in Laravels models is just an integer in the database.
So it has no affect.

MrKrisKrisu: These are the current values of our productive database: 
image

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
MrKrisKrisu: Ah, I see! Then the first instruction makes sense. However, I would still delete the second one.

app/Http/Controllers/StatusController.php
HerrLevin: #249

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
HerrLevin: >  So it has no affect.

*effect. But yeah, I guess. I'll change that accordingly.

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
HerrLevin: Now it should be better.

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public function down() {
+        Schema::table('statuses', function(Blueprint $table) {
+            $table->boolean('business')
+                ->change();
+        });
+
+        DB::table('statuses')
+            ->where('business', '>=', '1')
```
MrKrisKrisu:
```suggestion
            ->where('business', '>', '1')
```

MrKrisKrisu (changes requested):

database/migrations/2021_04_21_102707_change_statuses_to_accept_business_checkin_as_integer.php
MrKrisKrisu: Thanks!

MrKrisKrisu (approved):

PR #276, opened 2021-05-02
🩹 remove option to follow yourself via the api

### LINKED ISSUE #268: You can follow yourself
You can follow yourself

jeyemwey:
Closes #268. Since the `UserController/createFollow` only returns a boolean, I had to modify the translation files (thanks deepl for sv).


MrKrisKrisu:
Thanks for the PR! I'm glad to see you here again. :)

Can you check again for the failed test? I retriggered it again, but it continues to fail.

About the "problem" with the boolean: maybe you can work with exceptions there? For example "UserAlreadyFollowException" or something like that? So you can deliver more specific messages.

Otherwise it would be nonblocking for me.

MrKrisKrisu:
Since you edited the file: I have always wondered what the Babel file is for. Can you explain it to me?

MrKrisKrisu (approved):

Issue #278, opened 2021-05-04
Clicking on 'Home' icon cause an error


EventHorizon8:
**Describe the bug**
There is an error, when I try to click on 'Home' icon.

**To Reproduce**
1. Go to 'Dashboard'
2. Search by 'Göttingen'
3. Click on 'Home' icon

**Expected behavior**
 'Göttingen' saved as your home address.

**Screenshots**
<img width="647" alt="Screenshot 2021-05-04 at 12 37 03" src="https://user-images.githubusercontent.com/24681580/116985818-ab7c6580-acd5-11eb-8f69-75a07d0020d8.png">

**Additional context**
This error appeared because of front-end request parameters.

PR #279, opened 2021-05-04
#278 Work with $station as an object. Change id to idnr in frontend

EventHorizon8:
Fix a bug #278, which I noticed.


aledjones (approved):
Thank you for the fix!

PR #281, opened 2021-05-04
#273 fix express filter and regional filter

### LINKED ISSUE #273: Filtering by trip type does not seem to work.
Filtering by trip type does not seem to work. The error has yet to be confirmed.

image

EventHorizon8:
there is only one filter in app 'express' which gather  'nationalExpress' and 'express' from api, I guess there is the same situation with 'regionalExp' and 'regional' types


MrKrisKrisu:
Thanks for helping out! I'll approve and merge this PR. 

The tests need to be adapted to this change, I will do this in another PR.

MrKrisKrisu (approved):

EventHorizon8:
@MrKrisKrisu Maybe it will be better to use some Enums instead of travelTypes values? It will be easier to use constants in tests and other parts of code.

MrKrisKrisu:
Yes, you are right that this would be a better option. :)

PR #333, opened 2021-05-12
⚡️ Mute users

### LINKED ISSUE #323: Mute users
I don't want to see a user's posts, so I want to be able to block/mute them.

Add a button to mute users so that they don't appear in my dashboard.

/

**Additional context**
This feature can be implemented with the following logic components:
- New Table for "muted users"/"blocked users" structured like follows/follow requests
- New Table/page in settings to manage blocked/muted users
- Extend logic of `getUserInvisibleToMeAttribute` in `User.php` to accommodate for the blocked/muted users
- If this is implemented as a real block-function, the blocked users should not be able to follow me, see my statuses, like my statuses. This part would need a lot more functionality. I'd recommend to only implement muting and creating a new issue for blocking.

MrKrisKrisu:
closes #323


app/Http/Controllers/Backend/UserController.php
```diff
@@ -0,0 +1,47 @@
...
+use App\Http\Controllers\Controller;
+use App\Models\User;
+use App\Models\UserMute;
+use Exception;
+
+class UserController extends Controller
+{
+
+    /**
+     * @throws UserAlreadyMutedException
```
HerrLevin, line 19: This phpdoc-comment is incomplete.

app/Http/Controllers/Backend/UserController.php
```diff
@@ -0,0 +1,47 @@
...
+            $user->load('mutedUsers');
+            return true;
+        } catch (Exception $exception) {
+            report($exception);
+            return false;
+        }
+    }
+
+    /**
+     * @throws UserNotMutedException
```
HerrLevin, line 42: s.o.

app/Http/Controllers/Frontend/UserController.php
```diff
@@ -0,0 +1,60 @@
...
+    public function muteUser(Request $request): RedirectResponse {
+        $validated = $request->validate([
+                                            'user_id' => [
+                                                'required',
+                                                'exists:users,id',
+                                                Rule::notIn(auth()->user()->mutedUsers->pluck('id'))
+                                            ]
+                                        ]);
+
+        $userToBeMuted = User::find($validated['user_id']);
```
HerrLevin, line 26: not findorfail? or is this handled by the try/catch down below?

HerrLevin (changes requested):

app/Http/Controllers/Backend/UserController.php
MrKrisKrisu: parameter types and return value is already documented in the body. A description is unnecessary in my opinion, since the function is self-describing, so I would not add it either.

app/Http/Controllers/Backend/UserController.php
MrKrisKrisu: parameter types and return value is already documented in the body. A description is unnecessary in my opinion, since the function is self-describing, so I would not add it either.

app/Http/Controllers/Frontend/UserController.php
MrKrisKrisu: It's handled in line 21.

app/Http/Controllers/Backend/UserController.php
HerrLevin: It's not about how well the function explains itself. The parameters `User $user, User $userToBeMuted` should be in the phpdoc-comment so that every IDE can give the correct context for a method. PHPstorm has an easy generator for that.

app/Http/Controllers/Backend/UserController.php
HerrLevin: It's not about how well the function explains itself. The parameters User $user, User $userToBeMuted should be in the phpdoc-comment so that every IDE can give the correct context for a method. PHPstorm has an easy generator for that.

HerrLevin (approved):

PR #344, opened 2021-05-15
🐛 Fixed likes

### LINKED ISSUE #348: Status cannot be liked
**Describe the bug**
If I press the like button on any status, nothing happens. Instead a error message is put out on the JS browser console:
image

**To Reproduce**
1. Go to Dashboard page
2. Press the fav-button on any status

**Expected behavior**
HTTP request for liking the state will be sent and the star's state will change

MrKrisKrisu:


HerrLevin (changes requested):
better practice would be to change the data-tag in the view to `data-trwl-statusid`

MrKrisKrisu:
> better practice would be to change the data-tag in the view to `data-trwl-statusid`

I don't understand?

HerrLevin:
> I don't understand?

I started using `data-trwl-XXX` in the HTML instead of `data-XXX` to better distinguish our data-attributes from those of other things (just like bootstrap/mdb does with `data-md-xxx`/`data-bs-xxx`). It would be better to change the name of the data-attributes in the HTML to the newer standard than changing the js back to the old one. 

For reference: `data-trwl-asdf` would translate to `data.trwlAsdf`in JS.

HerrLevin (approved):

PR #347, opened 2021-05-17
♻️ moved social URL generation into user model

HerrLevin:


app/Http/Controllers/UserController.php
```diff
@@ -161,44 +161,8 @@ public static function getProfilePage($username): ?array {
...
-                    $getInfo    = $connection->get('users/show', ['user_id' => $user->socialProfile->twitter_id]);
-                    $twitterUrl = "https://twitter.com/" . $getInfo->screen_name;
-                } catch (Exception $e) {
-                    // The big whale time or $user has removed the api rights but has not told us yet.
-                }
-            }
-        }
-
+        $twitterUrl  = $user->twitterUrl;
+        $mastodonUrl = $user->mastodonUrl;
```
MrKrisKrisu: Good Idea! However, I would now still remove these two lines from the controller and instead simply access the user variable in the blade. This saves another 4 lines.

MrKrisKrisu (changes requested):

app/Http/Controllers/UserController.php
HerrLevin: I even reduced the amount of variable declarations by just moving the attribute getters to the array.

MrKrisKrisu (approved):

PR #350, opened 2021-05-19
🙈 Removed mix-manifest.json

MrKrisKrisu:


MrKrisKrisu:
Ah, lol. I'm stupid.

MrKrisKrisu:
Hmm, now the tests are failing, because the pre-generated manifest isn't there... But I don't really want to add `npm run` to the PHPUnit test...

.gitignore
```diff
@@ -3,6 +3,7 @@ node_modules/
 npm-debug.log
 report/
 clover.xml
+public/mix-manifest.json
```
HerrLevin, line 7: shouldn't there also be a `mix-manifest.json`?

HerrLevin (changes requested):

.gitignore
MrKrisKrisu: Yes, there should be a `mix-manifest.json`. But not in the repo as it is regenerated everytime `npm run x` is running.

HerrLevin (approved):

PR #358, opened 2021-05-21
⚡️ Improved leaderboard performance

MrKrisKrisu:


MrKrisKrisu:
I am puzzled as to why both of these tests fail here. When run locally, they work fine. I have also tested the API and frontend again manually. Can anyone help me with this? @HerrLevin @jeyemwey 

image
image

MrKrisKrisu:
The failing notification test is fixed in #357. The other one is not my fault and even failing in the develop. Should be fixed at another point.

HerrLevin (approved):

PR #383, opened 2021-05-28
🚑️ Fixed Leaderboard data and performance

### LINKED ISSUE #382: Leaderboard shows lifetime data
**Describe the bug**
The leaderboard shows the distance and duration of all check-ins.

**Expected behavior**
The leaderboard shows only the values of the last 7 days (for the main leaderboard) and the values of the respective month (monthly leaderboard).

I have already debugged the problem. The calculated values are not overwritten correctly at the following:
https://github.com/Traewelling/traewelling/blob/e2b5b006d857f4a1c28cc700aca24ac8f1621025/app/Http/Controllers/Backend/LeaderboardController.php#L79-L80

**Screenshots**
image
download

MrKrisKrisu:
The leaderboard now shows the correct data again (last 7 days, or selected month). Also, it is now 27x faster than before this change.

The API v1 endpoint needs to be adjusted.


resources/views/leaderboard/month.blade.php
```diff
@@ -16,13 +16,13 @@
                 <hr/>
                 <a href="{{route('leaderboard.month', ['date' => $date->clone()->subMonth()->format('Y-m')])}}"
                    class="btn btn-sm btn-primary float-left">
-                    <i class="fas fa-arrow-left"></i> {{$date->clone()->subMonth()->isoFormat(__('dateformat.month-and-year'))}}
+                    <em class="fas fa-arrow-left"></em> {{$date->clone()->subMonth()->isoFormat(__('dateformat.month-and-year'))}}
```
HerrLevin: please change <em> tags back to <i> tags with a `aria-hidden=true`-Tag.

HerrLevin (changes requested):

resources/views/leaderboard/month.blade.php
```diff
@@ -21,8 +21,8 @@ class="btn btn-sm btn-primary float-left">
                 @if($date->clone()->addMonth()->isBefore(\Carbon\Carbon::now()->endOfMonth()))
                     <a href="{{route('leaderboard.month', ['date' => $date->clone()->addMonth()->format('Y-m')])}}"
                        class="btn btn-sm btn-primary float-end">
-                        {{$date->clone()->addMonth()->isoFormat(__('dateformat.month-and-year'))}} <i
-                                class="fas fa-arrow-right"></i>
+                        {{$date->clone()->addMonth()->isoFormat(__('dateformat.month-and-year'))}}
+                        <i class="fas fa-arrow-right"></i>
```
HerrLevin, line 25: they're still missing the `aria-hidden="true`

HerrLevin (approved):

resources/views/leaderboard/month.blade.php
MrKrisKrisu: I don't want to change this here because the PR is about a different topic.

Issue #340, opened 2021-05-13
Fix PrivateProfileVisibilityTest


MrKrisKrisu:
**Describe the bug**
Often the PrivateProfileVisibilityTest fails without reason. After retriggering the test, it no longer fails. Please correct this or delete the test if it does not serve its purpose. 

**Expected behavior**
The test does not fail if it does not contain an error

**Screenshots**
![image](https://user-images.githubusercontent.com/4103693/118110431-aa88b980-b3e2-11eb-8b43-6749815727c3.png)
![image](https://user-images.githubusercontent.com/4103693/118110397-a3fa4200-b3e2-11eb-89ab-1596b2f490f3.png)

PR #384, opened 2021-05-28
✅🙄 Skip broken tests

MrKrisKrisu:
See #340


HerrLevin (approved):
I mean... that's  also one way to do it. 😂

PR #393, opened 2021-06-06
🚸 Hide past events without check-ins

### LINKED ISSUE #391: Hide past events if nobody checked in
Hide past events if nobody checked in

MrKrisKrisu:


app/Http/Controllers/FrontendStatusController.php
```diff
@@ -139,6 +139,10 @@ public function getActiveStatuses(): Renderable {
     public function statusesByEvent(string $event): Renderable {
         $response = StatusController::getStatusesByEvent($event, null);
 
+        if($response['event']->end->isPast() && $response['statuses']->count() == 0) {
+            abort(404);
+        }
+
```
jeyemwey, line 145: I'm not sure if a `404 Not Found` response is accurate here. If someone has the URL to the event already, the 404 error page would be disturbing to them.

I'd suggest to remove all traces of the event, but not delete the event itself?

MrKrisKrisu: It was my intention to delete the event completely, since it is not needed by anyone.

HerrLevin (approved):

PR #394, opened 2021-06-06
🐛 Prevent users from muting themselves

### LINKED ISSUE #355: You can mute yourself
**Describe the bug**
You can mute yourself.

**Expected behavior**
You can't mute yourself.

MrKrisKrisu:


app/Http/Controllers/Frontend/UserController.php
```diff
@@ -19,7 +19,8 @@ public function muteUser(Request $request): RedirectResponse {
                                             'user_id' => [
                                                 'required',
                                                 'exists:users,id',
-                                                Rule::notIn(auth()->user()->mutedUsers->pluck('id'))
+                                                Rule::notIn(auth()->user()->mutedUsers->pluck('id')),
+                                                Rule::notIn([auth()->user()->id]),
```
jeyemwey, line 23: Why is the second rule called with an array? Could both lines be merged in one rule that's called with an array?

resources/views/profile.blade.php
```diff
@@ -25,25 +25,8 @@
...
-                                @csrf
-                                <input type="hidden" name="user_id" value="{{$user->id}}"/>
-                                <button type="submit" class="btn btn-sm btn-primary" data-mdb-toggle="tooltip"
-                                        title="{{ __('user.mute-tooltip') }}">
-                                    <i class="far fa-eye-slash"></i>
-                                </button>
-                            </form>
+                        @if(auth()->user()->id != $user->id)
+                            @include('includes.mute-button')
                         @endif
```
jeyemwey: Can we move the gate to `resources/views/includes/mute-button.blade.php` too, please? That would be parallel to `resources/views/includes/follow-button.blade.php`.

app/Http/Controllers/Frontend/UserController.php
MrKrisKrisu: Because "notIn" expects an array of values that are not allowed. We can create an array and out all values in it but I think this would be more messy as these solution

jeyemwey (approved):

PR #405, opened 2021-06-07
🚸 Force user to reload styles and scripts if needed

MrKrisKrisu:


HerrLevin (commented):
Interesting approach. But... should we really?

jeyemwey:
How long does each `get_current_git_commit()` take? Because it [will constantly read from a file](https://github.com/Traewelling/traewelling/blob/724e305d39d88d4fcffe6c45e9c611752f1af1bc/app/Helpers/Helper.php#L10) in a blocking behavior (because well, PHP) which isn‘t particularly „free“. Maybe add a cache to it, so the file is only read once?

HerrLevin:
shouldn't the `mix('app.css')` method replicate exactly this behavior? (I'm not sure. I'm mostly guessing. Can't imagine any other use)

**Edit**: Yes, it should. The laravel docs should be sufficient info for that: [here](https://laravel.com/docs/8.x/mix#versioning-and-cache-busting)

MrKrisKrisu:
> Interesting approach. But... should we really?

I currently manually set the current timestamp behind the link for each release. I would be happy if this works automatically. I don't care how it is done. This would be my suggestion.

MrKrisKrisu:
> How long does each `get_current_git_commit()` take? Because it [will constantly read from a file](https://github.com/Traewelling/traewelling/blob/724e305d39d88d4fcffe6c45e9c611752f1af1bc/app/Helpers/Helper.php#L10) in a blocking behavior (because well, PHP) which isn‘t particularly „free“. Maybe add a cache to it, so the file is only read once?

Good point! I will have a look at this tomorrow.

MrKrisKrisu:
> shouldn't the `mix('app.css')` method replicate exactly this behavior? (I'm not sure. I'm mostly guessing. Can't imagine any other use)
> 
> **Edit**: Yes, it should. The laravel docs should be sufficient info for that: [here](https://laravel.com/docs/8.x/mix#versioning-and-cache-busting)

Oh nice! Will change this tomorrow.

HerrLevin (approved):

PR #407, opened 2021-06-07
🐛 Fix distance calculation

### LINKED ISSUE #10: 0km bus ride because saarVV is weird
**Describe the bug**
https://traewelling.de/status/118639 is a 0km bus ride, because the RMV is doing weird things with Hafas.

**To Reproduce**
1. Search for `Homburg Hbf`
2. Select an R7 bus to `ZOB (Stadtmitte), Zweibrücken`
3. Check-in until `ZOB (Stadtmitte), Zweibrücken`.
4. Check the logged distance 

**Expected behavior**
The distance is correct (~ 13km).

**Screenshots**
grafik

**Additional context**

RMV seems to squeeze two busses (one-way and back) into a single hafas trip (see attached screenshot from the API). While this does not _explain_ the error on this particular journey, it certainly contributes to its weirdness.

MrKrisKrisu:


MrKrisKrisu:
fixes #10 
image

MrKrisKrisu:
It works in principle. But it would probably make sense to write a few tests before merging this.

MrKrisKrisu:
This trip is now also longer than just 0 km.
image

HerrLevin (approved):

PR #438, opened 2021-06-16
✨ Vue-CheckIns

HerrLevin:


app/Http/Controllers/API/v1/TransportController.php
```diff
@@ -72,4 +77,74 @@ public function getTrip(Request $request): JsonResponse {
...
+        $validator = Validator::make($request->all(), [
+            'body'        => 'max:280',
+            'business'    => Rule::in(Business::getList()),
+            'visibility'  => Rule::in(StatusVisibility::getList()),
+            'event'       => 'integer',
+            'tweet'       => 'boolean',
+            'toot'        => 'boolean',
+            'ibnr'        => 'boolean',
+            'tripID'      => 'required',
+            'lineName'    => 'required',
```
MrKrisKrisu: Can you define whether all the things are `nullable` or `required`? Just to have a good overview.

app/Http/Controllers/API/v1/TransportController.php
```diff
@@ -72,4 +77,74 @@ public function getTrip(Request $request): JsonResponse {
...
         return $this->sendv1Response(data: $trainTripResponse);
     }
+
+
+    public function create(Request $request): JsonResponse {
+        $validator = Validator::make($request->all(), [
+            'body'        => 'max:280',
+            'business'    => Rule::in(Business::getList()),
+            'visibility'  => Rule::in(StatusVisibility::getList()),
+            'event'       => 'integer',
```
MrKrisKrisu: Can you add the rule `exists:events,id` to event?

MrKrisKrisu: Can you change event to something like eventID?

app/Http/Controllers/API/v1/TransportController.php
```diff
@@ -72,4 +77,74 @@ public function getTrip(Request $request): JsonResponse {
...
+            'tweet'       => 'boolean',
+            'toot'        => 'boolean',
+            'ibnr'        => 'boolean',
```
MrKrisKrisu: Why is `ibnr` boolean?

app/Http/Controllers/StatusController.php
```diff
@@ -418,4 +419,32 @@ public static function getFutureCheckins(): Paginator {
                          $query->where('departure', '>=', date('Y-m-d H:i:s', strtotime("+20min")));
                      })->simplePaginate(15);
     }
+
+    public static function createStatus(
+        User $user,
+        bool $business,
+        $visibility,
```
MrKrisKrisu, line 426: Is there a reason why the type of $visibility is not specified?

app/Http/Controllers/StatusController.php
```diff
@@ -418,4 +419,32 @@ public static function getFutureCheckins(): Paginator {
...
+        string $body = null,
+        int $eventId = null,
+        string $type = "hafas"
+    ): Status {
+        $event = null;
+        if ($eventId !== null) {
+            $event = Event::find($eventId);
+            //FixMe: Is the "new Carbon" really needed here?
+            if (!Carbon::now()->isBetween(new Carbon($event?->begin), new Carbon($event?->end))) {
```
MrKrisKrisu: new Carbon should not be needed, as begin and end should be an carbon Object already

app/Http/Controllers/TransportController.php
```diff
@@ -431,10 +435,89 @@ public static function TrainCheckin($tripId,
...
+     * @api v1
+     */
+    public static function createTrainCheckin(
+        Status $status,
+        HafasTrip $trip,
+        $entryStop,
+        $exitStop,
+        Carbon $departure = null,
+        Carbon $arrival = null,
+        bool $ibnr = false): array {
```
MrKrisKrisu: the formatting is a little bit confusing. Can you add an nl after `false`?

app/Http/Controllers/TransportController.php
```diff
@@ -431,10 +435,89 @@ public static function TrainCheckin($tripId,
...
+     * @throws StationNotOnTripException
+     * @throws CheckInCollisionException
+     * @throws ModelNotFoundException
+     * @api v1
+     */
+    public static function createTrainCheckin(
+        Status $status,
+        HafasTrip $trip,
+        $entryStop,
+        $exitStop,
```
MrKrisKrisu: can you specify the types?

app/Http/Controllers/TransportController.php
```diff
@@ -431,10 +435,89 @@ public static function TrainCheckin($tripId,
...
+
+        if (!$ibnr) {
+            $firstStop = $trip->stopoversNEW->where('train_station_id', $entryStop)
+                                            ->where('departure_planned', $departure)->first();
+
+            $lastStop = $trip->stopoversNEW->where('train_station_id', $exitStop)
+                                           ->where('arrival_planned', $arrival)->first();
+        } else {
+            //will this work @kris?
+            $firstStop = $trip->stopoversNEW->where('train_station_id.ibnr', $entryStop)
```
MrKrisKrisu: I don't think the train_station_**id** has an ibnr? :D

app/Http/Controllers/TransportController.php
```diff
@@ -431,10 +435,89 @@ public static function TrainCheckin($tripId,
...
+                                                 'delay'       => $trip->delay,
+                                                 'points'      => $points,
+                                                 'departure'   => $firstStop->departure_planned,
+                                                 'arrival'     => $lastStop->arrival_planned
+                                             ]);
+        foreach ($trainCheckin->alsoOnThisConnection as $otherStatus) {
+            $otherStatus->user->notify(new UserJoinedConnection(statusId: $status->id,
+                linename: $trip->linename,
+                origin: $firstStop->name,
+                destination: $lastStop->name));
```
MrKrisKrisu: Can you improve the formatting? Looks a little bit confusing

MrKrisKrisu (changes requested):

app/Http/Controllers/API/v1/TransportController.php
HerrLevin: So that external tools (like travelynx) still can use Träwelling, since the IBNR is a standard in germany

app/Http/Controllers/StatusController.php
HerrLevin: Yes, because with the current enum-types this will fail. Especially b/c we can't use typecasting.

app/Http/Controllers/API/v1/TransportController.php
```diff
@@ -81,13 +81,13 @@ public function getTrip(Request $request): JsonResponse {
...
-            'visibility'  => Rule::in(StatusVisibility::getList()),
-            'event'       => 'integer',
-            'tweet'       => 'boolean',
-            'toot'        => 'boolean',
-            'ibnr'        => 'boolean',
+            'body'        => ['nullable', 'max:280'],
+            'business'    => ['nullable', Rule::in(Business::getList())],
+            'visibility'  => ['nullable', Rule::in(StatusVisibility::getList())],
+            'eventID'       => ['nullable', 'integer', 'exists:events,id'],
+            'tweet'       => ['nullable', 'boolean'],
```
MrKrisKrisu: format please :)

MrKrisKrisu (changes requested):

app/Http/Controllers/StatusController.php
MrKrisKrisu: simply use int for this? Or am I wrong?

HerrLevin: I'm not sure about that since the enums aren't fully int and this would be a bit stupid to exclude them. :D

app/Http/Controllers/API/v1/TransportController.php
```diff
@@ -72,4 +77,74 @@ public function getTrip(Request $request): JsonResponse {
...
+            'destination' => ['required', 'numeric'],
+            'departure'   => ['required', 'date'],
+            'arrival'     => ['required', 'date'],
+        ]);
+        if ($validator->fails()) {
+            return $this->sendError($validator->errors(), 400);
+        }
+
+        try {
+            $hafasTrip = HafasController::getHafasTrip($request->input('tripID'), $request->input('lineName'));
```
MrKrisKrisu: is hafasTrip used?

MrKrisKrisu (changes requested):

MrKrisKrisu (approved):

PR #465, opened 2021-07-24
🕒🎉 Carbon-Party

MrKrisKrisu:


app/Http/Middleware/PrivacyInterceptionMiddleware.php
```diff
@@ -29,22 +27,25 @@ public function handle($request, Closure $next) {
             return $next($request);
         }
 
-        // Wenn die letzte Ausführung neuer ist als das Ack, redirecte mich bitte.
-        if ($user->privacy_ack_at <= $agreement->valid_at) {
+        // If the last execution is newer than the ack, please redirect me.
+        $user = auth()->user();
+        if ($user->privacy_ack_at == null || $agreement->valid_at->isAfter($user->privacy_ack_at)) {
```
HerrLevin: could be shortened to `if ($agreement->valid_at->isAfter($user->privacy_ack_at)) {`

Please test first though

resources/views/privacy-interception.blade.php
```diff
@@ -21,15 +21,15 @@
...
+                    @elseif($user->privacy_ack_at->isBefore($agreement->valid_at))
                         <div class="card mb-3">
                             <p class="card-body mb-0">
                                 {!! __('privacy.we-changed') !!}
                             </p>
                         </div>
                     @endif
 
-                    @if($user->privacy_ack_at < $agreement->valid_at || $user->privacy_ack_at == null)
+                    @if($user->privacy_ack_at == null || $agreement->valid_at->isAfter($user->privacy_ack_at))
```
HerrLevin: Could also be shortened to `@if($agreement->valid_at->isAfter($user->privacy_ack_at))`

Again test first please

HerrLevin (changes requested):

resources/views/privacy-interception.blade.php
MrKrisKrisu: Nope, because privacy_ack_at is null at default.

app/Http/Middleware/PrivacyInterceptionMiddleware.php
MrKrisKrisu: Nope, because privacy_ack_at is null at default.

HerrLevin (approved):

PR #466, opened 2021-07-24
⬆️ Update abraham/twitteroauth to v3

MrKrisKrisu:
...and update other dependencies


app/Http/Controllers/Backend/Social/TwitterController.php
```diff
@@ -0,0 +1,28 @@
...
+abstract class TwitterController extends Controller
+{
+    /**
+     * @param User $user
+     * @return TwitterOAuth
+     * @throws InvalidArgumentException
+     */
+    public static function getApi(User $user): TwitterOAuth {
+        if ($user->socialProfile->twitter_token == null || $user->socialProfile->twitter_tokenSecret == null) {
+            throw new InvalidArgumentException('User is not connected to twitter');
```
HerrLevin: You should catch this exception and handle it accordingly. Currently this error would be logged which seems to be a bit unnecessary. Also Either the type of the exception or the message should be changed to sth more fitting. e.g. "Empty Tokens were provided for user"

HerrLevin (changes requested):

app/Http/Controllers/TransportController.php
```diff
@@ -616,66 +615,6 @@ public static function createTrainCheckin(
...
-    /**
-     * @param Status $status
-     */
-    public static function postTwitter(Status $status): void {
-        if (config('trwl.post_social') !== true) {
-            return;
-        }
-
-        try {
-            $connection = TwitterController::getApi($status->user);
```
HerrLevin: You accidentally reverted your changes here.

HerrLevin (changes requested):
You accidentally reverted your changes to the `postTwitter` Method

HerrLevin (approved):

PR #467, opened 2021-07-24
⚡️ Replaced twitter link by unique link

MrKrisKrisu:
If the ToDo in the Vue templates was also implemented it saves a query to Twitter on every page load.

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -170,6 +170,9 @@ class UserController extends Controller
                     })->orderByDesc('created_at')->paginate(15);
     }
 
+    /**
+     * @todo remove twitterUrl after implemented ID-Link in vue Template
+     */
     public static function getProfilePage($username): ?array {
         $user = User::where('username', 'like', $username)->first();
         if ($user === null) {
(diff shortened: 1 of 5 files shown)


MrKrisKrisu:
@todo for @HerrLevin 
Maybe you can do this in this PR before it is merged

HerrLevin:
> @todo for @HerrLevin
> Maybe you can do this in this PR before it is merged

Sorry, but why didn't you just change the `getTwitterUrlAttribute()` to return this link? This would've changed it in vue and in in blade. 😅

app/Http/Controllers/UserController.php
```diff
@@ -170,6 +170,9 @@ public static function statusesForUser(User $user): ?LengthAwarePaginator {
                     })->orderByDesc('created_at')->paginate(15);
     }
 
+    /**
+     * @todo remove twitterUrl after implemented ID-Link in vue Template
```
HerrLevin, line 174: See https://github.com/Traewelling/traewelling/pull/467#issuecomment-886096709

HerrLevin (changes requested):

app/Http/Controllers/UserController.php
MrKrisKrisu: Don't you think the method would be unnecessary if it just returned a string without doing anything?

HerrLevin: Hm. Maybe.

HerrLevin (approved):

PR #468, opened 2021-07-24
📝 Updated the documentation of the .env and of the setup section in readme

HerrLevin:


.env.example
```diff
@@ -6,47 +7,58 @@ APP_URL=http://localhost
...
 
-# The derhuerst/db-rest:v5 instance which we use.
-# You can create your own instance
-DB_REST=http://localhost:3000
+# This is the official derhuerst/db-rest:v5 instance. Please set up your own if you're planning on using this more than
+# once in a while so their servers are not being overloaded.
+DB_REST=https://v5.db.transport.rest
 
-# Prevent TRWL from posting to socialMedia
+# When set to TRUE, Träwelling will not attempt to post to social media, even if the users want to.
```
MrKrisKrisu:
```suggestion
# When set to FALSE, Träwelling will not attempt to post to social media, even if the users want to.
```
:D

.env.example
```diff
@@ -6,47 +7,58 @@ APP_URL=http://localhost
...
 
+# Mastodon app credentials, only needed for logging in with and posting to mastodon
+# https://docs.joinmastodon.org/methods/apps/
 MASTODON_DOMAIN=
 MASTODON_ID=
 MASTODON_SECRET=
 MASTODON_REDIRECT=
 MASTODON_APPNAME=
 
+# Default values to be shown in the imprint. Especially important in Germany.
```
MrKrisKrisu:
```suggestion
# Default values to be shown in the legal notice.
```
I think it is not only important in germany. also "imprint" is not the best wording. I know, we are using it as a title there, but we should change this in the future to "legal notice" or something.

README.md
```diff
@@ -24,18 +24,19 @@
...
-* A database of choice, e.g. MariaDB or SQLite (preferably MariaDB)
-	* If you're using SQLite, make sure you have `php-sqlite` installed
+To set up a Träwelling instance you'll need:
+* [npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
+* [MariaDB](https://mariadb.org/download) (SQLite is used for running tests)
 * A local instance of [db-rest v5](https://github.com/derhuerst/db-rest/tree/5)
-
-After setting up those things, you can clone the repository and get the dependencies:
+* [Composer](https://getcomposer.org/download/)
+* PHP 8.0 (or higher) and the following extensions:
```
MrKrisKrisu:
```suggestion
* PHP 8.0 and the following extensions:
```
We currently only support ^8.0

MrKrisKrisu (changes requested):

README.md
HerrLevin: But....  8.0.8!

.env.example
HerrLevin: I think as long as the value is called "imprint_XXX" we should be consistent here.

MrKrisKrisu (approved):

PR #469, opened 2021-07-24
🔧 Added default run configurations for phpstorm

HerrLevin:


MrKrisKrisu (approved):
I don't know what the files are used for. But since they don't affect the project in a big way I will trust you.

PR #491, opened 2021-08-09
🚑 Private users check ins should now be hidden in the global dashboard.

HerrLevin:


app/Http/Controllers/StatusController.php
```diff
@@ -150,9 +151,9 @@ public static function getGlobalDashboard(): Paginator {
                                ->orWhere('users.id', $user)
                                ->orWhere(function($query) {
                                    $followings = Auth::check() ? auth()->user()->follows()->select('follow_id') : [];
-                                   $query->where('visibility', StatusVisibility::FOLLOWERS)
-                                         ->whereIn('users.id', $followings)
-                                         ->orWhere('visibility', StatusVisibility::PUBLIC);
+                                   $query->whereIn('users.id', $followings)
+                                         ->where('visibility', StatusVisibility::PUBLIC)
+                                         ->where('visibility', StatusVisibility::FOLLOWERS);
```
MrKrisKrisu: This cannot work?

MrKrisKrisu (changes requested):

app/Http/Controllers/StatusController.php
HerrLevin: Why not?

MrKrisKrisu: If visibility is public AND followers?

HerrLevin: Ohhh. So maybe 
```php
->whereNot('visibility', StatusVisibility::Private)
->whereNot('visibility', StatusVisibility::Unlisted)
```

Or  a 
```php
->where(function ($query) {
    $query->where('visibility', StatusVisibility::Public)
          ->orWhere('visibility', SatusVisibility::Followers)
});
```
?
What would be the better solution?

MrKrisKrisu:
## easy to understand

select `statuses`.*, (select count(*) from `likes` where `statuses`.`id` = `likes`.`status_id`) as `likes_count` from `statuses` inner join `train_checkins` on `train_checkins`.`status_id` = `statuses`.`id` inner join `users` on `statuses`.`user_id` = `users`.`id` where ((`users`.`private_profile` = ? and `visibility` = ?) or `users`.`id` = ? or (`users`.`id` in (select `follow_id` from `users` inner join `follows` on `users`.`id` = `follows`.`follow_id` where `follows`.`user_id` = ?) and `visibility` not in (?, ?))) and `train_checkins`.`departure` < ? and `statuses`.`user_id` not in (select `muted_id` from `users` inner join `user_mutes` on `users`.`id` = `user_mutes`.`muted_id` where `user_mutes`.`user_id` = ?) order by `train_checkins`.`departure` desc

MrKrisKrisu (approved):

PR #493, opened 2021-08-14
✨ Implement Follow-Button in Vue.JS

HerrLevin:


app/Exceptions/IdentidalModelException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace App\Exceptions;
+
+use Exception;
+
+class IdentidalModelException extends Exception
```
MrKrisKrisu: you mean identical?
```suggestion
class IdenticalModelException extends Exception
```

HerrLevin: Oops

MrKrisKrisu (approved):

PR #518, opened 2021-09-25
🚸 Require follower approval without private profile

MrKrisKrisu:
UserStory: I want to allow only certain followers, but leave my profile public. In the meantime, I can set individually per status who is allowed to see it.


app/Http/Controllers/Frontend/SettingsController.php
```diff
@@ -48,17 +48,28 @@ public function updateMainSettings(Request $request): RedirectResponse {
...
                                         ]);
 
+        $privateProfile   = isset($validated['private_profile']) && $validated['private_profile'] == 'on';
+        $followerApproval = isset($validated['follower_approval']) && $validated['follower_approval'] == 'on';
+
+        if ($privateProfile) {
+            $followerApproval = true;
+        }
+        if (!$followerApproval) {
+            $privateProfile = false;
```
HerrLevin, line 72: I'm not sure if that's a good idea. Wouldn't that mean that a private profile would become public if you uncheck the followerApproval?

resources/views/includes/follow-button.blade.php
```diff
@@ -1,21 +1,25 @@
 @if($user->id == Auth::user()->id)
     <a href="{{ route('settings') }}" class="btn btn-sm btn-primary">{{ __('profile.settings') }}</a>
-@elseif($user->private_profile && $user->followRequests->contains('user_id', Auth::user()->id))
+@elseif(($user->private_profile || $user->follower_approval) && $user->followRequests->contains('user_id', Auth::user()->id))
```
HerrLevin, line 3: might be better if we introduced an attribute that generally evaluates whether follows of a user should be approved.

resources/views/includes/follow-button.blade.php
```diff
@@ -1,21 +1,25 @@
...
     @endif
 @endif
 <script>
-    window.translFollow = "{{__('profile.follow')}}";
+    window.translFollow   = "{{__('profile.follow')}}";
     window.translUnfollow = "{{__('profile.unfollow')}}";
-    window.translPending = "{{__('profile.follow_req.pending')}}";
-</script>
\ No newline at end of file
+    window.translPending  = "{{__('profile.follow_req.pending')}}";
```
HerrLevin: nothing to do with your PR, I just hate this hacky JS workaround.

MrKrisKrisu: same.

PR #528, opened 2021-10-02
🌐 Add validation translations

MrKrisKrisu:
@copyright https://github.com/Laravel-Lang/lang MIT-License


MrKrisKrisu:
This is just a workaround.

Including the package would be too much. We can consider whether to include individual translations in Weblate. But we have to discuss that.

HerrLevin (approved):

PR #533, opened 2021-10-02
🎨 Rewrite and optimize exports

MrKrisKrisu:


HerrLevin:
Is it really necessary to add a csvexportcontroller & a jsonexportcontroller? Isn't a general exportcontroller preferrable?

MrKrisKrisu:
> Is it really necessary to add a csvexportcontroller & a jsonexportcontroller? Isn't a general exportcontroller preferrable?

In each export type we export different data. Therefore I think this is the best way.

resources/views/pdf/export-template.blade.php
```diff
@@ -104,20 +104,29 @@
                     <span class="page-number">{{ __('export.page') }}</span>
                 </div>
                 <div class="center">
-                    1 {{ __('export.reason.business') }} | 2 {{ __('export.reason.commute') }}
+                    0 {{ __('export.reason.private') }} | 1 {{ __('export.reason.business') }} | 2 {{ __('export.reason.commute') }}
                 </div>
                 <div class="left">
-                    <span class="promo">{!! __('export.guarantee', ['url' => url('/'), 'name' => config('app.name', 'Träwelling')]) !!}</span>
+                    <span
```
HerrLevin: reformat (newline after > instead of span

HerrLevin (approved):

HerrLevin (approved):

PR #539, opened 2021-10-08
✨ Implement social media login via API

Vogeslu:
Idea: Allow users to gain an access token just by logging in with twitter (or later mastodon) by providing an additional query parameter in the redirect route used for normal authentications with twitter and mastodon.

When a request is done with the parameter ``http://localhost:8000/auth/redirect/twitter?return=token`` the response is a json with the user token and expire date, see below

image
(key is local, so... useless here)

PS: I don't have mastodon and trouble testing it, so i added a small todo, propably the same procedure as in the SocialController.php

### DIFF
--- a/app/Http/Controllers/SocialController.php
+++ b/app/Http/Controllers/SocialController.php
@@ -10,6 +10,7 @@ use Exception;
 use File;
 use GuzzleHttp\Exception\ClientException;
 use Illuminate\Database\QueryException;
+use Illuminate\Http\JsonResponse;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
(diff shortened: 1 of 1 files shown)


aledjones:
Forgive if I'm blind, but I don't see a use case for this feature. Can you give me an example where you would need this kind of behavior?

MrKrisKrisu:
> Forgive if I'm blind, but I don't see a use case for this feature. Can you give me an example where you would need this kind of behavior?

If you add ?return=token to the Twitter login link, you can get a traewelling API token after the user confirms the login. This can be useful for our community apps.

aledjones:
So... to enable login via socials in community apps? 🤔

aledjones:
Ohhhhh, I think I get it now. Thanks for pointing that out 🙈

MrKrisKrisu:
I will merge this into `dev-ks-539` and add it to our API structure. Thanks!

PR #563, opened 2021-10-15
💄 Drop AdminLTE, add Bootstrap 5 for AdminPanel

MrKrisKrisu:


resources/js/admin.js
```diff
@@ -1,10 +1,10 @@
...
 
 window.addEventListener("load", () => {
     require("./components/usageBoard");
 });
 
-// For the event tabs:
-require("bootstrap/js/dist/tab");
\ No newline at end of file
+window.Popper = require("@popperjs/core");
+window.$      = window.jQuery = require("jquery");
```
HerrLevin: do we really still need jquery in the admin panel even though you're updating to bootstrap 5?

MrKrisKrisu: I do not know any place where it would required

PR #564, opened 2021-10-17
✨ Add user search to vue + api

HerrLevin:


MrKrisKrisu:
Please change the base branch from `main` to `develop`.

app/Http/Controllers/Backend/UserController.php
```diff
@@ -57,4 +59,23 @@ public static function unmuteUser(User $user, User $userToBeUnmuted): bool {
...
     }
+
+    /**
+     * @param string|null $searchQuery
+     *
+     * @return Paginator
+     * @throws InvalidArgumentException
+     */
+    public static function searchUser(?string $searchQuery): Paginator {
+        $validator = Validator::make(['searchQuery' => $searchQuery], ['searchQuery' => 'required|alpha_num']);
```
MrKrisKrisu:
```suggestion
        $validator = Validator::make(['searchQuery' => $searchQuery], ['searchQuery' => ['required', 'alpha_num']]);
```

app/Http/Controllers/Backend/UserController.php
```diff
@@ -57,4 +59,23 @@ public static function unmuteUser(User $user, User $userToBeUnmuted): bool {
...
+        if ($validator->fails()) {
+            throw new InvalidArgumentException();
+        }
+
+        return User::where(
+            'name', 'like', "%{$searchQuery}%"
+        )->orWhere(
+            'username', 'like', "%{$searchQuery}%"
+        )->simplePaginate(10);
```
MrKrisKrisu, line 79: I don't like Paginator as a return type. Can you change this to return either an Collection or just the query?

MrKrisKrisu (changes requested):

app/Http/Controllers/Backend/UserController.php
HerrLevin: And how do we paginate the entire response? 😅

MrKrisKrisu: One way is to just remove the `->simplePaginate(10);` here and return the raw non-executed query. You can use the simplePaginate in your API Frontend and if another code-block needs the result in this function in another way it can be easily used. But a paginate in the backend is too limited.

MrKrisKrisu (approved):

PR #569, opened 2021-10-17
✨ Add error popups

HerrLevin:


resources/components/views/Charts.vue
```diff
@@ -358,7 +359,12 @@ export default {
                     this.updateVolume();
                 })
                 .catch((error) => {
-                    console.error(error);
+                    this.loading = false;
+                    if (error.response) {
+                        this.notyf.error(error.response.data.message);
+                    } else {
+                        this.notyf.error(this.i18n.get("_.messages.exception.general"));
+                    }
```
MrKrisKrisu: Can you write one function for this and include it on every step instead of duplicate code?

resources/components/views/Charts.vue
```diff
@@ -370,7 +376,12 @@ export default {
                     this.fromGlobal  = response.data.meta.from;
                 })
                 .catch((error) => {
-                    console.error(error);
+                    this.loading = false;
+                    if (error.response) {
+                        this.notyf.error(error.response.data.message);
+                    } else {
+                        this.notyf.error(this.i18n.get("_.messages.exception.general"));
+                    }
```
MrKrisKrisu: Can you write one function for this and include it on every step instead of duplicate code?

MrKrisKrisu (changes requested):

MrKrisKrisu (approved):

PR #583, opened 2021-10-26
🥅 Add a timeout error to HAFAS API and return errors correctly

HerrLevin:


.env.testing
```diff
@@ -23,6 +23,7 @@ MAIL_PASSWORD=52e9e5002d9db7
 MAIL_ENCRYPTION=tls
 
 DB_REST=https://v5.db.transport.rest/
+DB_REST_TIMEOUT=3
```
MrKrisKrisu: Please have a look at the tests. DB Rest is working now, so it shouldnt fail

MrKrisKrisu (changes requested):

MrKrisKrisu (approved):

PR #585, opened 2021-10-27
🚀 Docker image for easier deployment

aledjones:
- [x] Dockerfile: creates image with prebuilt webpack
- [x] Dockerfile: has maintainer labels
- [x] Dockerfile: use composer image
- [x] docker-entrypoint.sh: migrates current database on start
- [ ] docker-entrypoint.sh: locking mechanism for database migration
- [ ] docker-entrypoint.sh: error-handling
- [ ] docker-entrypoint.sh: locking state validation
- [x] GitHub Actions: Images are built on push to defined branches
- [x] GitHub Actions: Images are built and tagged with commit id
- [x] GitHub Actions: Images are pushed to GitHub Registry tagged with the gut reference that started the build (branch push or tag release)


Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM node:16 as NodeBuildContainer
+COPY . /usr/src/trwl
+WORKDIR /usr/src/trwl
+RUN npm i
+RUN npm run prod
+
+FROM php as PHPBuildContainer
```
MrKrisKrisu: Can you specify PHP 8.0?

app/Providers/AppServiceProvider.php
```diff
@@ -26,6 +27,9 @@ public function register(): void {
      * @throws BindingResolutionException
      */
     public function boot(): void {
+        if (env('APP_ENV') === 'prod') {
```
MrKrisKrisu:
```suggestion
        if (config('app.env') === 'prod') {
```

docker-entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -e
+
+if [ "$1" = 'apache2-foreground' ]; then
+    wait-for-it $DB_HOST:${DB_PORT:=3306}
+    cd /var/www/html
+    runuser -u www-data -- php artisan migrate
```
MrKrisKrisu:
```suggestion
    runuser -u www-data -- php artisan migrate --force
```

Isn't this required?

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" && \
+    php composer-setup.php --install-dir=/bin --filename=composer
+RUN apt update && apt install -y zlib1g-dev libpng-dev zip
+RUN docker-php-ext-install gd exif
```
MrKrisKrisu: We have already talked about that. In the long run, we should use a composer image and set a version. I have written it on the todo. But it is non-blocking for me.

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+COPY --from=NodeBuildContainer /usr/src/trwl /usr/src/trwl
+WORKDIR /usr/src/trwl
+RUN composer install
+
+FROM php:apache
```
MrKrisKrisu: Can you specify PHP 8.0?

.htaccess
```diff
@@ -0,0 +1,6 @@
+Options +FollowSymLinks
+RewriteEngine On
+
+RewriteCond %{REQUEST_FILENAME} !-d
+RewriteCond %{REQUEST_FILENAME} !-f
+RewriteRule ^ index.php [L]
```
MrKrisKrisu: What is the file needed for? It is located in the root directory.

In the public folder I would understand it.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,48 @@
+# shamelessly copied from https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#upgrading-a-workflow-that-accesses-ghcrio
+# modified to meet our needs
+name: "Docker Image"
+
+on:
+  push:
+    branches:
+      - dev-*
```
MrKrisKrisu: I would not build a container on our dev-* branches. Or do you think that makes sense?

.dockerignore
```diff
@@ -0,0 +1,3 @@
+.git
```
MrKrisKrisu, line 1: maybe also add vendor/ and node_modules? This would at least be useful for local building.

MrKrisKrisu (changes requested):

.dockerignore
aledjones: Will do

.github/workflows/docker.yml
aledjones: This was only a suggestion for easier local and on-premise testing like in cloud environments...
I'll remove those entries as they can be added back in the future

aledjones: On second thought: this is the only way to test docker builds in my PR branch.
It'll stay and be removed as soon as we merge into develop.

.htaccess
aledjones: This file just shows that I'm an idiot that didn't put together a laravel container for a very long time.
Will be removed in upcoming changes.

Dockerfile
aledjones: Will do.

Dockerfile
aledjones: Changes to `composer:2` in upcoming commit

app/Providers/AppServiceProvider.php
```diff
@@ -26,6 +27,9 @@ public function register(): void {
      * @throws BindingResolutionException
      */
     public function boot(): void {
+        if (config('app.env') === 'prod') {
```
MrKrisKrisu:
```suggestion
        if (config('app.env') === 'production') {
```
Sorry, I missed that last time. xD

MrKrisKrisu (changes requested):

MrKrisKrisu (approved):

PR #598, opened 2021-11-03
🧹🔥 Sort structure to frontend/backend, remove deprecated code

MrKrisKrisu:


app/Http/Controllers/Backend/User/DashboardController.php
```diff
@@ -0,0 +1,69 @@
...
+                     ->orderBy('train_checkins.departure', 'desc')
+                     ->whereIn('statuses.user_id', $followingIDs)
+                     ->whereIn('visibility', [StatusVisibility::PUBLIC, StatusVisibility::FOLLOWERS])
+                     ->orWhere('statuses.user_id', $user->id)
+                     ->withCount('likes')
+                     ->latest()
+                     ->simplePaginate(15);
+    }
+
+    public static function getGlobalDashboard(User $user): Paginator {
```
HerrLevin, line 36: Do we really need to supply a user here? Isn't a default `auth()->user()` enough?

app/Http/Controllers/Backend/User/DashboardController.php
```diff
@@ -0,0 +1,69 @@
...
+use App\Models\Status;
+use App\Models\User;
+use Carbon\Carbon;
+use Illuminate\Contracts\Pagination\Paginator;
+use Illuminate\Database\Eloquent\Builder;
+
+abstract class DashboardController extends Controller
+{
+
+    public static function getPrivateDashboard(User $user): Paginator {
```
HerrLevin, line 16: Do we really need to supply a user here? Isn't a default `auth()->user()` enough?

app/Http/Controllers/FrontendStatusController.php
```diff
@@ -27,7 +28,7 @@ class FrontendStatusController extends Controller
 {
     public function getDashboard(): Renderable|RedirectResponse {
         $user     = Auth::user();
-        $statuses = StatusBackend::getDashboard($user);
+        $statuses = DashboardController::getPrivateDashboard($user);
```
HerrLevin: `DashboardController::getPrivateDashboard(auth()->user());`

app/Http/Controllers/Backend/User/DashboardController.php
```diff
@@ -34,19 +30,16 @@ public static function getGlobalDashboard(): Paginator {
...
-                             $query->orWhere('users.id', auth()->user()->id);
-                         }
+                         $query->orWhere('users.id', $user->id);
 
                          //Option 3: Status is from a followed BUT not unlisted or private
-                         $query->orWhere(function(Builder $query) {
-                             $followings = Auth::check() ? auth()->user()->follows()->select('follow_id') : [];
-                             $query->whereIn('users.id', $followings)
+                         $query->orWhere(function(Builder $query) use ($user) {
+                             $query->whereIn('users.id', $user->follows()->select('follow_id'))
```
HerrLevin, line 58: This needs discussion since it removes the ability to access the global dashboard unauthenticated.

MrKrisKrisu: Is this a Thing? I didn't know this ist possible.

HerrLevin (changes requested):
See previous comments.

app/Http/Controllers/Backend/User/DashboardController.php
MrKrisKrisu: Each user has a different global dashboard based on private profiles and their follows. I think it makes sense to differentiate explicitly.

app/Http/Controllers/Backend/User/DashboardController.php
MrKrisKrisu: Each user has a different private dashboard based on private profiles and their follows. I think it makes sense to differentiate explicitly. In addition, you can theoretically test better this way.

app/Http/Controllers/Backend/User/DashboardController.php
MrKrisKrisu: In addition, you can theoretically test better this way.

app/Http/Controllers/Backend/User/DashboardController.php
MrKrisKrisu: In addition, you can theoretically test better this way.

app/Http/Controllers/API/v1/TransportController.php
```diff
@@ -174,16 +178,21 @@ public function create(Request $request): JsonResponse {
      */
     public function setHome(string $stationName): JsonResponse {
         try {
-            $station = TransportBackend::setTrainHome(user: auth()->user(), stationName: $stationName);
+            $trainStation = HafasController::getStations(query: $stationName, results: 1)->first();
+            if ($trainStation === null) {
+                return $this->sendv1Error("Your query matches no station", 404);
```
HerrLevin: code is redundant, default is 404.

HerrLevin:
```suggestion
                return $this->sendv1Error("Your query matches no station");
```

HerrLevin (changes requested):

HerrLevin (approved):

PR #606, opened 2021-11-05
Reindex departure array after sort

### LINKED ISSUE #584: API v1 Wrong array type is returned for departures
**Describe the bug**
In some cases when i want to fetch the departures for a station the data is an object instead of an array (Example: ``/api/v1/trains/station/8010308/departures?when=2021-10-27T15:52:20.172Z``)

Normal Behavior:

```
{
    "data": [
        {
            "tripId": "1|1819680|1|80|27102021",
            // ...
        }
    ],
    "meta": {
        // ...
    }
}
```

Odd Behavior:

```
{
    "data": {
        "1": {
            "tripId": "1|218226|0|80|27102021",
            // ...
        },
        "0": {
            "tripId": "1|1401662|2|80|27102021",
            // ...
        }
    },
    "meta": {
        // ...
    }
}
```

**To Reproduce**

Open ``/api/v1/trains/station/8010308/departures?when=2021-10-27T15:52:20.172Z``

jheubuch:
fixes #584

When sorting the departure array by the departure/plannedDeparture, the keys in the collection get mixed up.

Because of this, the output JSON is encoded as an object instead of an array. The collection keys needs to be reindexed so that it can be encoded correctly.

### DIFF
--- a/app/Http/Controllers/TransportController.php
+++ b/app/Http/Controllers/TransportController.php
@@ -98,7 +98,7 @@ class TransportController extends Controller
             return $departure->when ?? $departure->plannedWhen;
         });
 
-        return ['station' => $station, 'departures' => $departures, 'times' => $times];
+        return ['station' => $station, 'departures' => $departures->values(), 'times' => $times];
     }
 
     /**


MrKrisKrisu:
Thanks!

PR #608, opened 2021-11-06
🚸 Improve station search

MrKrisKrisu:
Changes:

1. enable station search by name, DS100 and (new!) ID from TrainStation.
2. remove stationLink function, as it is no longer needed. Hafas can now handle the differences.
3. When clicking on a station, the ID is used by default to search uniquely. When searching by station name, there were problems especially with smaller stations. (see internal ticket #145661)


app/Http/Controllers/Backend/Transport/StationController.php
```diff
@@ -0,0 +1,43 @@
...
+{
+
+    /**
+     * @throws HafasException
+     * @throws ModelNotFoundException
+     */
+    public static function lookupStation(string|int $query): TrainStation {
+        //Lookup by station id
+        if (is_numeric($query)) {
+            $station = TrainStation::find($query);
```
HerrLevin: This seems like a bad solution to me. This will completely restrict to a) internal IDs and b) only cached stations.

Maybe we should think about using a flag for numeric values to distinguish between ids and ibnr?

resources/views/activejourneys.blade.php
```diff
@@ -244,7 +244,7 @@ className: 'text-trwl text-center'
                                 "end": "{{ $event->end->format('Y-m-d') }}",
                                 "ts": {!! $event->getTrainStation() !!},
                                 "mapLink": "{{ route('statuses.byEvent', ['eventSlug' => $event->slug]) }}",
-                                "closestLink": `{!! stationLink($event->getTrainstation()->name) !!}`
+                                "closestLink": `<a href="{{route('trains.stationboard', ['provider' => 'train', 'station' => $event->getTrainstation()->id])}}" class="text-trwl clearfix">{{$event->getTrainstation()->name}}</a>`
```
HerrLevin:
```suggestion
                                "closestLink": "<a href=\"{{route('trains.stationboard', ['provider' => 'train', 'station' => $event->getTrainstation()->id])}}\" class=\"text-trwl clearfix\">{{$event->getTrainstation()->name}}</a>"
```

Please check if this works. Might have made a typo

HerrLevin (changes requested):

HerrLevin (approved):

PR #641, opened 2021-12-05
🎅 Christmas time

MrKrisKrisu:
user


MrKrisKrisu:
I have made a reminder to myself that I will revert this after Christmas

MrKrisKrisu:
The snowflakes come only when they are explicitly activated by the user. And only for a short time for the session. Otherwise it would be annoying.

PR #808, opened 2022-04-03
:sparkles: Add minified api logging

MrKrisKrisu:
We need minified logging for API calls to know which API endpoints are used by which applications, so we can focus us on developing the most important ones. We don't save any user information, ip addresses or anything else. The user agent is required to identify applications.

We need this informations to know which routes from APIv0 can be shut down, as of APIv1 is currently in development.

image

### DIFF
--- a/app/Http/Controllers/Frontend/Admin/ApiUsageController.php
+++ b/app/Http/Controllers/Frontend/Admin/ApiUsageController.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace App\Http\Controllers\Frontend\Admin;
+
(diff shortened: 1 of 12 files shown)


MrKrisKrisu:
cc to @derf, @jheubuch, @Vogeslu

We will not shut down any routes that are in use. Before APIv1 becomes official and APIv0 is disabled you will receive additional information and will be contacted directly. But this will take longer. Just as information in advance.

HerrLevin (changes requested):
api.log -> kernel.php

HerrLevin (approved):

PR #827, opened 2022-04-07
🚑 Catch unset stations in distance calculation progress

HerrLevin:


jeyemwey:
Whats the real diff here besides that you moved the methods?

app/Http/Controllers/Backend/GeoController.php
```diff
@@ -159,6 +103,74 @@ public static function calculateDistanceBetweenCoordinates(
...
+     * @param HafasTrip $hafasTrip
+     *
+     * @return mixed
+     * @throws JsonException
+     */
+    private static function getPolylineWithTimestamps(HafasTrip $hafasTrip): stdClass {
+        $geoJsonObj = json_decode($hafasTrip->polyline->polyline, false, 512, JSON_THROW_ON_ERROR);
+        $stopovers  = $hafasTrip->stopoversNEW;
+
+        $stopovers = $stopovers->map(function($stopover) {
```
HerrLevin, line 149: added this map here

app/Http/Controllers/Backend/GeoController.php
```diff
@@ -159,6 +103,74 @@ public static function calculateDistanceBetweenCoordinates(
...
+        foreach ($geoJsonObj->features as $polylineFeature) {
+            if (!isset($polylineFeature->properties->id)) {
+                continue;
+            }
+
+            $stopover = $stopovers->where('trainStation.ibnr', $polylineFeature->properties->id)
+                                  ->where('passed', false)
+                                  ->first();
+
+            if (is_null($stopover)) {
```
HerrLevin, line 163: added this catch here

HerrLevin (commented):
sorry, phpstorm gone rogue.

app/Http/Controllers/Backend/GeoController.php
MrKrisKrisu: I don't see any relevanc for this map... you don't need `passed` to be set by default?

PR #847, opened 2022-04-15
:zap: Improve queries

MrKrisKrisu:
Only two (fast!) queries instead of n + 1 (n default = 10)


app/Http/Controllers/HafasController.php
```diff
@@ -34,7 +34,7 @@ public static function getTrainStationByRilIdentifier(string $rilIdentifier): ?T
         try {
             $client   = new Client(['base_uri' => config('trwl.db_rest'), 'timeout' => config('trwl.db_rest_timeout')]);
             $response = $client->get("/stations/$rilIdentifier");
-            $data     = json_decode($response->getBody()->getContents());
+            $data     = json_decode($response->getBody()->getContents(), false, 512, JSON_THROW_ON_ERROR);
```
MrKrisKrisu, line 37: Yes, Levin. This is catched below.

PR #934, opened 2022-06-19
:sparkles: Add only authenticated users visibility

NyCodeGHG:
This PR adds a new status visibility which allows users to only show their status to users which are logged in.
This is useful for users who may be a bit paranoid about random website visitors, but still want to participate publicly and let other people know when they are on the same train.


MrKrisKrisu:
Thanks for your work, good idea! :)
I will definitely use that too. Will test your changes soon once and review.

MrKrisKrisu (approved):
Everything perfect, thank you! Even with tests. :)

PR #1016, opened 2022-08-25
:boom: Fixed business statistics from API

### LINKED ISSUE #660: [APIv1] Changes to Statistics
In my opinion it would be useful to change the data format of the `api/v1/statistics` endpoint.

In the "purpose" Array, the business types are delivered as a string (e.g. "private", "commute", ...). Instead it would be more useful to deliver those with the correlating Enum ID which is also used for e.g. checking into a connection (private = 0, commute = 2 etc).

Also the duration is delivered as a string. It could be delivered as a number.

image

jheubuch:
fixes #660
image


MrKrisKrisu (approved):
In my opinion, that's okay. But @HerrLevin should merge this, because I don't know if that is breaking.

PR #1023, opened 2022-08-30
Display a hint if a user receives 1 point

### LINKED ISSUE #426: Inform user why he received only one point
If a user has received only one point because he checked in too early or too late, additionally display a warning explaining why this is so. Some users do not know that.

M4rcDev:
Further, improved the test to use a fixed static time to avoid side effects.

Resolves #426


MrKrisKrisu (approved):
Thanks! :)

PR #1058, opened 2022-09-19
Add queue monitor to get insights in jobs system

jeyemwey:
Uses [laravel-queue-monitor](https://github.com/Traewelling/Laravel-Queue-Monitor) instead of [Laravel Horizon](https://github.com/laravel/horizon) since its simpler, and doesn't rely on Redis. 

Here is how the system could look like:
screenshot_queue-monitor

~~Open issues (that could be mitigated by sending PRs to the package):~~
* ~~sub-second jobs are not displayed properly. I think this is a problem with the package that can be fixed through a package update later. The database holds the `time_elapsed` correctly though (see screenshot below), so we could still use the data to observe performance issues.~~
* ~~As far as I can tell, there is no autopurge of old messages (i.e. after two weeks). _Dadurch Gefahr des Datenbank Vollscheißens 💩 ._ We could build this manually as another scheduled job or [add the scheduled job to the package](https://laracasts.com/discuss/channels/laravel/task-scheduling-in-packages?page=1&replyId=181270).~~

image

This will obviously get better once we make other tasks async (see Discord discussion).

### DIFF
--- a/app/Jobs/SendVerificationEmail.php
+++ b/app/Jobs/SendVerificationEmail.php
@@ -26,8 +26,7 @@ class SendVerificationEmail implements ShouldQueue
      *
      * @return void
      */
-    public function __construct(User $user)
-    {
+    public function __construct(User $user) {
         $this->user = $user;
     }
 
@@ -36,10 +35,9 @@ class SendVerificationEmail implements ShouldQueue
      *
      * @return void
      */
-    public function handle()
-    {
+    public function handle(): void {
         $this->queueData([
-                             "user_id" => $this->user->id,
+                             "user_id"  => $this->user->id,
                              "username" => $this->user->username,
                          ]);
 
--- a/app/Jobs/SendVerificationEmail.php
+++ b/app/Jobs/SendVerificationEmail.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace App\Jobs;
+
+use App\Models\User;
+use Illuminate\Auth\Notifications\VerifyEmail;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+use Traewelling\QueueMonitor\Traits\IsMonitored;
+
+/**
+ * Send the Email which verifies a user account asynchronously.
+ * @see https://aregsar.com/blog/2020/how-to-queue-laravel-user-verification-email/
+ */
+class SendVerificationEmail implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, IsMonitored, Queueable, SerializesModels;
+
+    protected User $user;
+
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct(User $user)
+    {
+        $this->user = $user;
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $this->queueData([
+                             "user_id" => $this->user->id,
+                             "username" => $this->user->username,
+                         ]);
+
+        // This queued job sends
+        // Illuminate\Auth\Notifications\VerifyEmail verification
+        // to the user by triggering the notification
+        $this->user->notify(new VerifyEmail);
+    }
+}
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -3,6 +3,7 @@
 namespace App\Models;
 
 use App\Enum\StatusVisibility;
+use App\Jobs\SendVerificationEmail;
 use Carbon\Carbon;
 use Exception;
 use Illuminate\Contracts\Auth\MustVerifyEmail;
@@ -234,5 +235,9 @@ class User extends Authenticatable implements MustVerifyEmail
      */
     public function notifications(): MorphMany {
         return $this->morphMany(DatabaseNotification::class, 'notifiable')->orderBy('created_at', 'desc');
+    }
+
+    public function sendEmailVerificationNotification() {
+        SendVerificationEmail::dispatch($this);
     }
 }
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -237,7 +237,7 @@ class User extends Authenticatable implements MustVerifyEmail
         return $this->morphMany(DatabaseNotification::class, 'notifiable')->orderBy('created_at', 'desc');
     }
 
-    public function sendEmailVerificationNotification() {
+    public function sendEmailVerificationNotification(): void {
         SendVerificationEmail::dispatch($this);
     }
 }
--- a/config/queue-monitor.php
+++ b/config/queue-monitor.php
@@ -1,0 +1,55 @@
+<?php
+
+return [
+    /*
+     * Set the table to be used for monitoring data.
+     */
+    'table' => 'queue_monitor',
+    'connection' => null,
+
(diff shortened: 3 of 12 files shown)


jeyemwey:
I've now forked the package to our organization, and added a number of features to the thing (i.e. cleaning out old monitoring tasks).

This whole PR now adds the following things:
1. A new dependency that keeps track of running jobs.
2. A scheduled command that removes old logs that are not visited anymore.
3. A new UI that shows the data. The UI is accessible to admin users under `/admin/queue-monitor` and is linked in the admin dashboard.
4. The given `FetchCarriageSequence` job was adjusted to work with the queue monitor. The changes in commit 7a021948b74d36d555a52cdf482bb3d505ca728d give you an overview of the necessary changes to let something work asynchronously.
5. Sending the "Verify your Email Address" email was moved to a job in 1f309539b1a83f13b28dad0af977a2e12c3d020a. This will reduce the loading time for this request, and we will learn issues earlier thanks to the advanced logging of the Jobs in the Queue Monitor.

@MrKrisKrisu Please take all the time that you need to review this. When shipping to production, please be aware that you need to run  `composer install` and `artisan:migrate` 🚀

M4rcDev:
I would consider to switch at some point to laravel horizion, although it requires redis.
Forking a package and maintaining it ourselves will cost us a lot of time in the long run.

The solution is fine for now, but please don't invest too much time in it.

jeyemwey:
>  to switch at some point to laravel horizion, although it requires redis.

I would not like to do that, since it requires every developer to run a redis instance. With this, the monitoring is one layer higher in the PHP code which allows me to make adjustments, and the developer can choose to use the sql db queuing without redis.

> Forking a package and maintaining it ourselves will cost us a lot of time in the long run.

Always depends on your usecase. IMO for us, this maintaining load is fine. Träwelling has already abstracted a number of features to packages (i.e. `mastodon-socialite`) and it's been fine.

> The solution is fine for now, but please don't invest too much time in it.

I'll be fine. ^^

MrKrisKrisu:
> Forking a package and maintaining it ourselves will cost us a lot of time in the long run.

I agree. I also look forward to the future somewhat with fear.

I have not yet dealt with monitoring in the laravel area. Horizon must necessarily use redis? In production I have no problem with that. You can set it up. But locally, hopefully you can work around that somehow?

But the PR is fine for me for now. This is currently not a major intervention and I would merge it immediately after review. But we can (and should) think about how we approach this in the long term...

> Träwelling has already abstracted a number of features to packages (i.e. mastodon-socialite) and it's been fine.

To this day, I still don't fully understand why this was done in the first place. Did the original package not work? I would also rather avoid.

app/Models/User.php
```diff
@@ -235,4 +236,8 @@ public function getMastodonUrlAttribute(): ?string {
     public function notifications(): MorphMany {
         return $this->morphMany(DatabaseNotification::class, 'notifiable')->orderBy('created_at', 'desc');
     }
+
+    public function sendEmailVerificationNotification() {
```
MrKrisKrisu:
```suggestion
    public function sendEmailVerificationNotification(): void {
```

app/Jobs/SendVerificationEmail.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->user = $user;
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
```
MrKrisKrisu:
```suggestion
    public function handle(): void
```

config/queue-monitor.php
```diff
@@ -0,0 +1,61 @@
+<?php
```
MrKrisKrisu, line 1: I would run the Autoformatter over the file.

app/Jobs/SendVerificationEmail.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
MrKrisKrisu, line 1: I would run the Autoformatter over the file.

database/migrations/2018_02_05_000000_create_queue_monitor_table.php
```diff
@@ -0,0 +1,43 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class CreateQueueMonitorTable extends Migration
+{
+    public function up()
+    {
+        Schema::create(config('queue-monitor.table'), function (Blueprint $table) {
+            $table->increments('id');
```
MrKrisKrisu:
```suggestion
            $table->id();
```
Can be simplified

database/migrations/2018_02_05_000000_create_queue_monitor_table.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class CreateQueueMonitorTable extends Migration
+{
+    public function up()
```
MrKrisKrisu:
```suggestion
    public function up(): void
```

database/migrations/2018_02_05_000000_create_queue_monitor_table.php
```diff
@@ -0,0 +1,43 @@
...
+
+            $table->longText('exception')->nullable();
+            $table->text('exception_message')->nullable();
+            $table->text('exception_class')->nullable();
+
+            $table->longText('data')->nullable();
+        });
+    }
+
+    public function down()
```
MrKrisKrisu:
```suggestion
    public function down(): void
```

resources/views/admin/layout.blade.php
```diff
@@ -66,7 +66,7 @@ class="nav-link text-white {{ request()->is('admin/users*') ? 'active' : '' }}">
                     <li>
                         <a href="{{ route('admin.locations') }}"
                            class="nav-link text-white {{ request()->is('admin/locations*') ? 'active' : '' }}">
-                            <i class="fa-solid fa-location-dot" aria-hidden="true"></i>
+                            <i class="fa-solid fa-map-location-dot me-2" aria-hidden="true"></i>
```
MrKrisKrisu, line 69: Thanks!

resources/views/vendor/queue-monitor/jobs.blade.php
```diff
@@ -0,0 +1,334 @@
...
+<html lang="en">
+
+<head>
+
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <title>@lang('Queue Monitor')</title>
+
+    <link href="https://unpkg.com/tailwindcss@^1.0/dist/tailwind.min.css" rel="stylesheet">
```
MrKrisKrisu, line 11: Please remove external css resources.
Maybe we can implement this in out Adminpanel layout?

MrKrisKrisu (changes requested):

jeyemwey:
Given the new information reg. monitoring, I think about removing our frontend again and using Grafana for this. That would make the un-fork possible and reduce complexity on our end.

Let's talk about this on Monday. I will try to check the complaints until then.

resources/views/vendor/queue-monitor/jobs.blade.php
jeyemwey: That is a rather large can of worms. I'd like to wait with that until we're sure that this frontend is used (instead of grafana).

jeyemwey:
Closing this PR in favor of

PR #1073, opened 2022-09-26
:bug: Fix checkin at entry/exit only stations

### LINKED ISSUE #1010: Check-in not possible when arrival time not set
**Describe the bug**
Some trains have stops which don't have an arrival time because they are not meant for getting off. However, you "can" still get off there. Currently traewelling shows the error "arrival muss ausgefüllt werden." when doing this. An example is shown below.

**To Reproduce**
Check in to ICE 1689 on 2022-08-14, From Hamburg Hbf (departs at 22:12) to Hamburg-Harburg. Harburg won't have an arrival time set.

**Expected behavior**
I think it should be possible to check in anyways because you "can" use this train to get from Hamburg Hbf to Harburg (I certainly did). If the arrival is blank it could simply be the same as the departure and vice-versa.

**Screenshots**
image

MrKrisKrisu:
fixes #1010


app/Http/Controllers/FrontendTransportController.php
```diff
@@ -143,8 +143,10 @@ public function TrainCheckin(Request $request): RedirectResponse {
         $validated = $request->validate([
                                             'tripID'            => ['required'],
                                             'start'             => ['required', 'numeric'], //Origin station IBNR
+                                            //if the train has no departure, arrival is okay
                                             'departure'         => ['required', 'date'],
                                             'destination'       => ['required', 'numeric'], //Destination station IBNR
+                                            //if the train has no arrival, departure is okay
                                             'arrival'           => ['required', 'date'],
```
jeyemwey: I don't understand yet, how this works, and how the comments change anything in the 'required' formula, but LGTM.

jeyemwey (approved):

app/Http/Controllers/FrontendTransportController.php
jeyemwey: @MrKrisKrisu Please advice, after that, I'd merge it.

MrKrisKrisu: I realize that these comments are more confusing than helpful. They only say that if no departure is available (e.g. because a train stops only to get off) the arrival is used as an alternative. I would remove the comments.

Issue #1035, opened 2022-09-07
Tweet Chaining


jeyemwey:
**Is your feature request related to a problem? Please describe.**
As a Long Distance Träweller, I'd like to chain all tweets in a single thread on journeys with multiple legs. This way, it is clearer that I'm on one long trip rather than on multiple short ones.

**Describe the solution you'd like**
During the checkin process, I'd like to have a toggle that enables this option.
Träwelling should then find my last checkin and post the new checkin text as a reply to that tweet.
If I have already started a thread off the first tweet (i.e. with pictures of the carriage), TRWL should append itself at the end of the chain.

**Describe alternatives you've considered**
As a low-effort alternative, the generated tweet text can be shown so I can copy and paste it at the end of the thread manually.

**Additional context / Solution Proposal**
This idea can contain multiple separate _stories_:
1. Save the tweet id / url next to some `train_checkin` information.
2. Implement logic to traverse to the end of the chain.
3. Allow the post logic to be a reply.
4. Deal with problems if the original tweet is deleted.
5. Create a checkbox on the checkin flow.

~NOTE: For the first iteration, this should only happen for Tweets, not Mastodon posts. Once the feature landed, we can recreate it in Mastodon later.~ Update(dec22): Mastodon Toot Chaining is implemented.

I'm happy to hear your thoughts about this!

M4rcDev:
That sounds good and I would use that feature as well. 👍🏼

MrKrisKrisu:
Good Idea. Maybe this overlaps a bit with #769.

jeyemwey:
I just spend my evening with the Twitter API. Because $things, the api doesn't give a great way to traverse a thread, but rather wants you to look through the answers individually. Given how all of this works, I'd like to mark this feature as `beta`, `try it out but no warranty` and make all wordings as cautious as possible.

That's how I believe the checkin flow could look like (you can download the PNG and upload it to draw.io:

![Diagramm](https://user-images.githubusercontent.com/2796271/194728334-0d4ce642-a9e3-487d-8bc2-a3161e3e1bd6.png)


And here's to a list of longer code snippets from the image:
* `$post_in_tweetchain && user()->statuses()->where("tweet_id", "!=", null)->exists()`
* `/tweets?ids= --- $status_origin->tweet_id --- &tweet.fields=conversation_id`
[...]

PR #1074, opened 2022-09-26
:monocle_face: Save tweet id for posted status

MrKrisKrisu:
preparation for #1035

### DIFF
--- a/app/Http/Controllers/Backend/Social/TwitterController.php
+++ b/app/Http/Controllers/Backend/Social/TwitterController.php
@@ -95,13 +95,14 @@ abstract class TwitterController extends Controller
                 $socialText .= " #dbl";
             }
             $socialText .= ' ' . url("/status/{$status->id}");
-            $connection->post("statuses/update",
-                              [
-                                  "status" => $socialText,
-                                  'lat'    => $status->trainCheckin->Origin->latitude,
-                                  'lon'    => $status->trainCheckin->Origin->longitude
-                              ]
+            $response   = $connection->post("statuses/update",
+                                            [
+                                                "status" => $socialText,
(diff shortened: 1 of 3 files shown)


app/Models/Status.php
```diff
@@ -32,6 +32,7 @@ class Status extends Model
         'business'   => Business::class,
         'visibility' => StatusVisibility::class,
         'event_id'   => 'integer',
+        'tweet_id'   => 'integer',
```
jeyemwey: I'm not sure how good PHP handles the uint_64 translation to integer. [Twitter recommends](https://developer.twitter.com/en/docs/twitter-ids) to use strings here to avoid [implicit conversions to Float](https://www.php.net/manual/en/language.types.integer.php) which are used in PHP.

app/Http/Controllers/Backend/Social/TwitterController.php
```diff
@@ -92,16 +92,17 @@ public static function postStatus(Status $status): void {
...
-                              ]
+            $socialText .= ' ' . url('/status/{$status->id}');
+            $response   = $connection->post('statuses/update',
+                                            [
+                                                'status' => $socialText,
+                                                'lat'    => $status->trainCheckin->Origin->latitude,
+                                                'lon'    => $status->trainCheckin->Origin->longitude
+                                            ]
             );
+            $status->update(['tweet_id' => $response->id]);
```
jeyemwey: I would expect this line under the 200 check (which should then `return` out). This way, we avoid an "unknown property" error if the http request turns out bad.

jeyemwey (changes requested):
Small changes, mostly nitpicking.

jeyemwey (approved):

PR #1088, opened 2022-10-07
:zap: Evaluate leaderboard cache every five minutes

jeyemwey:
This removes the burden of filling the leaderboard cache from the first user that visits the page every five minutes.

### DIFF
--- a/app/Console/Commands/CacheLeaderboard.php
+++ b/app/Console/Commands/CacheLeaderboard.php
@@ -15,7 +15,7 @@ class CacheLeaderboard extends Command
      *
      * @var string
      */
-    protected $signature = 'trwl:leaderboard:cache';
+    protected $signature = 'trwl:cache:leaderboard';
 
     /**
      * The console command description.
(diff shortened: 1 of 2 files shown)


app/Console/Commands/CacheLeaderboard.php
```diff
@@ -0,0 +1,41 @@
...
+use Illuminate\Support\Facades\DB;
+
+class CacheLeaderboard extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'trwl:leaderboard:cache';
```
MrKrisKrisu: I would switch the keys, so there is a better scheme, if we add another caching tasks later.

```suggestion
    protected $signature = 'trwl:cache:leaderboard';
```

MrKrisKrisu (changes requested):

MrKrisKrisu (approved):
But this is nbfm :)

PR #1091, opened 2022-10-09
:zap: Cache global stats

MrKrisKrisu:


jeyemwey:
I've added a number of changes in #1094 since I couldn't make those with Review Suggestions.

jeyemwey (approved):

PR #1093, opened 2022-10-09
:bug: Fix "fast check in" button not correctly selecting last stop if train end prematurely

DRSchlaubi:
Seemingly, db rest also cancels the last stop of a train; therefore we also check for departure/arrival for the trip station list but we don't for the fast check in

https://github.com/Traewelling/traewelling/blob/f3ab90f01ff09281293eb723302baaa0f7f6c833/resources/views/admin/checkin/trip.blade.php#L30

### DIFF
--- a/app/Http/Controllers/FrontendTransportController.php
+++ b/app/Http/Controllers/FrontendTransportController.php
@@ -121,7 +121,7 @@ class FrontendTransportController extends Controller
 
         // Find out where this train terminates and offer this as a "fast check-in" option.
         $terminalStopIndex = count($TrainTripResponse['stopovers']) - 1;
-        while ($terminalStopIndex >= 1 && @$TrainTripResponse['stopovers'][$terminalStopIndex]['cancelled'] == true) {
+        while ($terminalStopIndex >= 1 && @$this->is_cancelled($TrainTripResponse['stopovers'][$terminalStopIndex])) {
             $terminalStopIndex--;
         }
         $terminalStop = $TrainTripResponse['stopovers'][$terminalStopIndex];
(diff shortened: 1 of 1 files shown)


jeyemwey (approved):
Looks good to me. Thanks!

jeyemwey:
@MrKrisKrisu Somehow, the tests don't run through and thus I cannot merge it. Can you try that please?

If at all possible, I'd like to rebase the changes onto the current head to disguise the second metro line that comes from the earlier parent.

app/Http/Controllers/FrontendTransportController.php
```diff
@@ -121,7 +121,7 @@ public function TrainTrip(Request $request): Renderable|RedirectResponse {
 
         // Find out where this train terminates and offer this as a "fast check-in" option.
         $terminalStopIndex = count($TrainTripResponse['stopovers']) - 1;
-        while ($terminalStopIndex >= 1 && @$TrainTripResponse['stopovers'][$terminalStopIndex]['cancelled'] == true) {
+        while ($terminalStopIndex >= 1 && @$this->is_cancelled($TrainTripResponse['stopovers'][$terminalStopIndex])) {
```
MrKrisKrisu:
```suggestion
        while ($terminalStopIndex >= 1 && @$this->isCancelled($TrainTripResponse['stopovers'][$terminalStopIndex])) {
```

app/Http/Controllers/FrontendTransportController.php
```diff
@@ -225,4 +225,8 @@ public function setTrainHome(Request $request): RedirectResponse {
             return redirect()->back()->with(['error' => __('messages.exception.generalHafas')]);
         }
     }
+
+    private function is_cancelled(mixed $param): bool {
```
MrKrisKrisu:
```suggestion
    private function isCancelled(mixed $param): bool {
```

MrKrisKrisu (changes requested):

MrKrisKrisu:
> @MrKrisKrisu Somehow, the tests don't run through and thus I cannot merge it. Can you try that please?
> 
> If at all possible, I'd like to rebase the changes onto the current head to disguise the second metro line that comes from the earlier parent.

Normally we have a button to accept the action for "first time contributors".... But the button is missing. We can merge this anyway, because we don't have a test for this case.

PR #1173, opened 2022-11-10
:green_heart: Upgrade to actions/checkout@v3

jeyemwey:
See Deprecation Notice in https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/ and every CI run since then, all of which did post annotations.


MrKrisKrisu (approved):
Thanks for noticing! :)

PR #1187, opened 2022-11-20
:sparkles: Add sums to pdf export

jeyemwey:
This is my idea for the first part of #1174. There is no statistics regarding travel reasons, i don't know how we can implement that without kicking people who use the exporter for the Finanzamt.

image

Changes:
* Table Footer with summed numbers
* Duration and Distance are aligned to the right, and the unit will not wrap to the next line.
* Small Adjustments to the Reasons Footnote

Let me know what you think (:

### DIFF
--- a/app/Http/Controllers/Backend/Export/ExportController.php
+++ b/app/Http/Controllers/Backend/Export/ExportController.php
@@ -24,7 +24,16 @@ abstract class ExportController extends Controller
      * @throws DataOverflowException If too many results are given.
      */
     public static function getExportableStatuses(User $user, Carbon $timestampFrom, Carbon $timestampTo): Collection {
-        $statuses = Status::join('train_checkins', 'statuses.id', '=', 'train_checkins.status_id')
+        $statuses = Status::with([
+                                     'trainCheckin',
+                                     'trainCheckin.HafasTrip',
+                                     'trainCheckin.HafasTrip.stopoversNEW',
+                                     'trainCheckin.Origin',
+                                     'trainCheckin.Destination',
+                                     // 'trainCheckin.originStation',
+                                     // 'trainCheckin.destinationStation'
+                                 ])
+                          ->join('train_checkins', 'statuses.id', '=', 'train_checkins.status_id')
                           ->where('statuses.user_id', $user->id)
                           ->where('train_checkins.departure', '>=', $timestampFrom->startOfDay()->toIso8601String())
                           ->where('train_checkins.departure', '<=', $timestampTo->endOfDay()->toIso8601String())
@@ -89,8 +98,8 @@ abstract class ExportController extends Controller
             'statuses'     => $statuses,
             'begin'        => $begin,
             'end'          => $end,
-            'sum_duration' => $statuses->reduce(fn($sum, $s) => $sum + $s->trainCheckin->duration),
-            'sum_distance' => $statuses->reduce(fn($sum, $s) => $sum + $s->trainCheckin->distance) / 1000,
+            'sum_duration' => $statuses->sum('trainCheckin.duration'),
+            'sum_distance' => $statuses->sum('trainCheckin.distance') / 1000,
         ])
                   ->setPaper('a4', 'landscape')
                   ->download(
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Backend/Export/ExportController.php
```diff
@@ -83,10 +83,14 @@ private static function exportJson(Carbon $begin, Carbon $end): JsonResponse {
...
+
         return Pdf::loadView('pdf.export-template', [
-            'statuses' => ExportBackend::getExportableStatuses(auth()->user(), $begin, $end),
-            'begin'    => $begin,
-            'end'      => $end
+            'statuses'     => $statuses,
+            'begin'        => $begin,
+            'end'          => $end,
+            'sum_duration' => $statuses->reduce(fn($sum, $s) => $sum + $s->trainCheckin->duration),
+            'sum_distance' => $statuses->reduce(fn($sum, $s) => $sum + $s->trainCheckin->distance) / 1000,
```
MrKrisKrisu: I think this is more readable.

```suggestion
            'sum_duration' => $statuses->sum('trainCheckin.duration'),
            'sum_distance' => $statuses->sum('trainCheckin.distance') / 1000,
```

MrKrisKrisu (changes requested):

MrKrisKrisu:
> i don't know how we can implement that without kicking people who use the exporter for the Finanzamt.

You're right. That is a difficult question. I am also unsure if the statistics should be exportable as PDF here.
That's what the statistics page is for. You can also print them with the browser print function.

app/Http/Controllers/Backend/Export/ExportController.php
MrKrisKrisu: Maybe this can be also moved to the blade? In my opinion it is enough that we transfer `$statuses` from the Controller.

resources/views/pdf/export-template.blade.php
```diff
@@ -155,12 +162,25 @@
...
                     @endforeach
                 </tbody>
+                <tfoot>
+                    <tr>
+                        <td></td>
+                        <td></td>
+                        <td></td>
+                        <td></td>
+                        <td></td>
+                        <td style="font-style: italic;">Gesamt:</td>
```
MrKrisKrisu: Oh, I have now seen that the translation is missing here.

MrKrisKrisu (changes requested):

app/Http/Controllers/Backend/Export/ExportController.php
jeyemwey: Fixed with [dc482cc](https://github.com/Traewelling/traewelling/pull/1187/commits/dc482cc81292b04b0762c994575a929659035e36).

jeyemwey: I'd rather have that in the Controller, rather than in the blade.

resources/views/pdf/export-template.blade.php
jeyemwey: Fixed with [76e3c06](https://github.com/Traewelling/traewelling/pull/1187/commits/76e3c0693cd1045587641cf7b69002583439ddac).

app/Http/Controllers/Backend/Export/ExportController.php
```diff
@@ -24,7 +24,16 @@ abstract class ExportController extends Controller
      * @throws DataOverflowException If too many results are given.
      */
     public static function getExportableStatuses(User $user, Carbon $timestampFrom, Carbon $timestampTo): Collection {
-        $statuses = Status::join('train_checkins', 'statuses.id', '=', 'train_checkins.status_id')
+        $statuses = Status::with([
+                                     'trainCheckin',
+                                     'trainCheckin.HafasTrip',
```
MrKrisKrisu: This line can be removed because the third line also loads this relations. This saves 2 queries. :)
```suggestion
```

app/Http/Controllers/Backend/Export/ExportController.php
```diff
@@ -24,7 +24,16 @@ abstract class ExportController extends Controller
...
+                                     'trainCheckin.HafasTrip.stopoversNEW',
+                                     'trainCheckin.Origin',
+                                     'trainCheckin.Destination',
+                                     // 'trainCheckin.originStation',
+                                     // 'trainCheckin.destinationStation'
```
MrKrisKrisu: Also remove commented codeblocks - or do you need this in the future?
```suggestion
```

MrKrisKrisu (changes requested):

MrKrisKrisu (approved):
Thanks! :)

PR #1196, opened 2022-11-21
:technologist: Add default status visibility to /auth/user

DRSchlaubi:
This could be useful for apps like träweldroid to pre-select the default value, simply passing in "null" as visibility doesn't fully work, as the ui wouldn't know which visibility to pre-select

### DIFF
--- a/app/Http/Resources/UserBaseResource.php
+++ b/app/Http/Resources/UserBaseResource.php
@@ -33,11 +33,11 @@ class UserBaseResource extends JsonResource
             'preventIndex'    => $this->prevent_index,
             $this->mergeWhen(isset($this->UserSettingsResource),
                 [
-                    'role'          => $this->role,
-                    'home'          => $this->home,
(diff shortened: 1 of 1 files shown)


app/Http/Resources/UserBaseResource.php
```diff
@@ -36,7 +36,8 @@ public function toArray($request): array {
                     'role'          => $this->role,
                     'home'          => $this->home,
                     'dbl'           => $this->always_dbl,
-                    'language'      => $this->language
+                    'language'      => $this->language,
+                    'default_status_visibility' => $this->default_status_visibility
```
MrKrisKrisu:
```suggestion
                    'defaultStatusVisibility' => $this->default_status_visibility
```

MrKrisKrisu (changes requested):
Thanks! Can you also please align the variables? Then i'll merge the pr. :)

MrKrisKrisu (approved):

PR #1213, opened 2022-11-29
:zap: Improve CI performance by running tests in parallel

jeyemwey:
See https://blog.laravel.com/laravel-parallel-testing-is-now-available

PHP Tests [now take ~2:20](https://github.com/Traewelling/traewelling/actions/runs/3577663474/jobs/6016920961) instead of the 3:30-5:00 before. 🎉


MrKrisKrisu:
I removed this option for a better overview a few weeks/months(?) ago. But yeah, speed is more important I think.

MrKrisKrisu (approved):

Issue #425, opened 2021-06-11
Migrate old stopovers


MrKrisKrisu:
Database table: hafas_trips
Column: stopovers

- [x] Migrate old stopovers to the new train_stopovers table before removing!
- [x] Check, that this field is not required anywhere
- [x] Rename ->stopoversNEW relationship to ->stopovers

MrKrisKrisu:
After #1217 is merged the old stopovers column isn't used anymore. (I dont see any other requirements)

MrKrisKrisu:
After #1226 is merged there is only the migration of the old data required.

PR #1217, opened 2022-12-03
:wastebasket: Replace deprecated stopovers

MrKrisKrisu:
belongs to #425


resources/views/activejourneys.blade.php
```diff
@@ -38,7 +38,7 @@
                                 origin: {{$s->trainCheckin->origin}},
                                 destination: {{$s->trainCheckin->destination}},
                                 polyline: {!! $s->trainCheckin->hafastrip->polyline->polyline !!}, // Stored as JSON in DB
-                                stops: {!! $s->trainCheckin->hafastrip->stopovers !!}, // Stored as JSON in DB
+                                stops: {!! $s->trainCheckin->hafastrip->stopovers !!}, // Stored as JSON in DB //TODO: Replace ->stopovers with the new relation (see issue#425)
```
MrKrisKrisu: This is the only place where the obsolete stopovers are still used.

resources/js/components/ActiveStatus.js
```diff
@@ -0,0 +1,227 @@
+window.ActiveStatus = class ActiveStatus {
```
jeyemwey: Did you change something here or is this just the existing code, with other indentation?

And can you please rename the file to `activejourneys`, so its obvious that those two belong together?

MrKrisKrisu: No, currently there are no changes. I just moved the code to our webpack and put a wrapper class around it.
In the next step I will rewrite it to use the new stopovers table instead of the json.
Sure, I will rename it!

jeyemwey (approved):

PR #1222, opened 2022-12-08
:goal_net: Retry posting to social media with backoff

jeyemwey:
This would post the status later if the social media site is offline.

Backing off 10sec, 1min, 5min, 15min, 1h, 3h, 6h.


MrKrisKrisu (approved):
Good idea! But then we should also see whether we change the wording.
"I'm currently in ICE xxx" may no longer be relevant in 60 minutes.

But can me merged in my opinion. :)

PR #1253, opened 2022-12-25
:passport_control: Block users

jeyemwey:
The platform has evolved to need blocking behaviour. See #1251 for a recent example.

There is a block button on the profiles which fill and clear a `user_blocks` table 

When A blocks B (Brainstorming, open for discussions!!!):
- [x] The profile page of A is limited to B (no social links, no statistics, no check-ins)
- [x] The API returns a limited dataset or does not show A at all (error 403 or something)
- [x] To B, A shows the default profile image
- [x] Existing following relationships between A and B are removed (in both directions)
- [x] B can not see A's statuses on the `/status/{id}` page
- [x] B can not see A's statuses in the global dashboard and in `/statuses/active`
- [x] A can not see B's statuses in the global dashboard and in `/statuses/active`
- [ ] When A and B are in the same train, they both don't get a notification (see existing `todo` in `\App\Models\TrainCheckin::getAlsoOnThisConnectionAttribute`)
- [x] B's existing likes on A's statuses are not displayed or plainly deleted
- [x] In the `/settings/followers` page, a table of blocked accounts is shown
- [x] Tests describing the basic behaviour

Due to the holidays, I'm a little short on time. If _you as a reader_ find the time to work on any of these things, please run with the branch and continue working on it, so that our users can benefit from its features quickly.


jeyemwey:
IMO, this is ready for review. 

The open checkmark (`When A and B are in the same train, they both don't get a notification`) is also resolved, since the code below seems to work as intended. If people have blocked each other, they will not see that they're in the same train. I'm confused from the `todo` comment :/

image

app/Exceptions/NotAllowedException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace App\Exceptions;
+
+use Exception;
+
+class NotAllowedException extends Exception
```
MrKrisKrisu: Maybe just use `Illuminate\Auth\Access\AuthorizationException` instead?

MrKrisKrisu (approved):

PR #1260, opened 2022-12-27
:bug: Temporary remove eager loading

MrKrisKrisu:


MrKrisKrisu:
Some exports are broken. This belongs to a NullPointer where the `HafasTrip` can't be retrieved via the `TrainCheckin`. This error is a little bit confusing, because without the eager loading this is working fine. (but slower, because... eager loading is missing).

I'll merge this PR as a "hotfix".

PR #1263, opened 2022-12-29
Add stopovers to StatusResource.php

DRSchlaubi:
This would be useful to add the "Next station" feature to traewelldroid


HerrLevin:
@DRSchlaubi Hi, we already have an extra endpoint for stopovers: https://traewelling.de/api/v1/stopovers/714718 (multiple Statuses can be accessed by appending them with a `,` as a delimiter. 

We decided against stopovers inside the status model to reduce data consumption/speed up standard requests for dashboards, active statuses, etc.

PR #1295, opened 2023-01-13
:clown_face: Mock requests to db-rest

jeyemwey:
It is a [common issue](https://github.com/Traewelling/traewelling/actions/workflows/phpunit.yml?query=is%3Afailure) in our tests that they flap as db-rest is sometimes unavailable, or certain trains and stations have issues.

With this PR, most of the calls to db-rest are mocked and replaced with stub data. To achieve easier mocking, Laravel's HTTP client is used instead of Guzzle (see 5bc086f6442d02c3323d9de73bdf8ccc9bdd31e9).

Still missing: 
- [ ] `TestCase::checkin` is not mocked yet.
- [ ] After that, we can add `Http::preventStrayRequests();` to `TestCase::setUp`, so new tests will have to mock HTTP requests/responses.

Please don't be shocked by the size of the PR, >6900 lines come directly from JSON responses for the special tests (Bus 603, Ringbahn, Frankfurt Airport Bus).


jeyemwey:
@MrKrisKrisu I currently have no intentions to continue working on this PR and resolve the outstanding issues.

Since it offers direct benefit to the Tests in GitHub Actions (making them faster, reducing dependency on a "real" HAFAS), I suggest that this is merged today and we iterate over it afterwards.

There is a merge conflict though which I'd like you to resolve, since it directly impacts your changes in #1327. If you need help, please hit me up.

MrKrisKrisu:
After the merge something in `Transport/BackendCheckinTest.php` is throwing an general error, but I'm not sure how to handle this.

MrKrisKrisu (approved):

PR #1304, opened 2023-01-17
:sparkles: Implement Webhooks

### LINKED ISSUE #1301: Webhooks

When a third party application wants to know if a user did something on the platform (e.g. created a status, updated a status etc.), it currenlty must poll the data to do that.
Polling leads to high traffic.

Webhooks integrated into the existing system for third party applications.

I thought about creating api endpoints to create a webhook specific to a user.

Keep polling, but that would be bad for both third parties and traewelling because the traffic would be much higher here.

**Additional context**

I'm not that fluent in php but I would like to try to work on that feature. I just wanted to open this issue before i invest to much time into it and it won't get merged :)

NyCodeGHG:
See #1301


routes/api.php
```diff
@@ -123,6 +124,12 @@
             Route::get('follow-requests', [FollowController::class, 'getFollowRequests']);
             Route::get('followings', [FollowController::class, 'getFollowings']);
         });
+        Route::group(['prefix' => 'webhooks'], static function() {
+            Route::get('/', [WebhookController::class, 'getWebhooks']);
+            Route::post('/', [WebhookController::class, 'createWebhook']);
+            Route::delete('/{webhookId}', [WebhookController::class, 'deleteWebhook']);
+            Route::delete('/', [WebhookController::class, 'deleteAllWebhooks']);
```
jeyemwey: I suspect a `deleteAllWebhooks` route to be a footgun, especially if it isn't specified with an `all` keyword in the URL, but just by leaving the ID blank (although I understand that this is how REST works).

Why would someone use this endpoint? Please help me understand your thought process here :)

resources/views/settings/cards/webhooks.blade.php
```diff
@@ -0,0 +1,35 @@
...
+    <div class="card-body">
+        @if(count($webhooks) == 0)
+            <p class="text-danger">{{__('settings.no-access-from-external-applications')}}</p>
+        @else
+            <table class="table table-responsive">
+                <thead>
+                    <tr>
+                        <th>{{ __('settings.client-name') }}</th>
+                        <th>{{ __('settings.created') }}</th>
+                    </tr>
```
jeyemwey: Would it make sense to show the user to which events a webhook responds? Just thinking out lout here.

resources/lang/de.json
```diff
@@ -377,7 +377,7 @@
     "settings.prevent": "Verbieten",
     "settings.privacy.update.success": "Deine Privatsphäreneinstellungen wurden aktualisiert.",
     "settings.search-engines.description": "Wir setzen ein noindex-Tag auf deinem Profil um den Suchmaschinen mitzuteilen, dass dein Profil nicht indexiert werden darf. Wir haben keinen Einfluss darauf, ob die Suchmaschine diesen Wunsch berücksichtigt.",
-    "settings.no-tokens": "Es haben aktuell keine externen Anwendungen Zugriff auf deinen Account.",
+    "settings.no-access-from-external-applications": "Es haben aktuell keine externen Anwendungen Zugriff auf deinen Account.",
```
jeyemwey: Few things: If you change a key in the language file, please make sure that it isn't used elsewhere.

For the `no-tokens` key in particular, it's used in `resources/views/settings/cards/api-token.blade.php` where it makes sense as is.

Also, just because an application has not set a Webhook, it doesn't mean that it doesn't have access to the account, TRWL just does not push status updates to it.

In this case, I'd suggest duplicating the line and adding a `settings.no-webhooks` key, which could say something like "Es werden keine externen Anwendungen über Änderungen an deinem Account informiert". This way, no existing behaviour is modified.

app/Models/User.php
```diff
@@ -157,6 +157,10 @@ public function icsTokens(): HasMany {
         return $this->hasMany(IcsToken::class, 'user_id', 'id');
     }
 
+    public function webhooks(): HasMany {
+        return $this->hasMany(Webhook::class);
+    }
+
```
jeyemwey, line 161: Is this surfaced in the API when `/users/{id}` is called? 
Anyone, please resolve if this is the case.

app/Http/Controllers/Frontend/Transport/StatusController.php
```diff
@@ -37,6 +38,8 @@ public function updateStatus(Request $request): JsonResponse|RedirectResponse {
                                 'visibility' => StatusVisibility::from($validated['checkinVisibility']),
                             ]);
 
+            StatusUpdateEvent::dispatch($status->refresh());
```
jeyemwey, line 41: Does this StatusUpdateEvent take a changed destination in account? It _might_, but I'm not sure. When we write tests for this, this could be a special case to look out for.

routes/api.php
NyCodeGHG: It was useful for testing it locally, but yeah it's probably a footgun and we should remove it

resources/views/settings/cards/webhooks.blade.php
NyCodeGHG: It definitely would, good point

resources/lang/de.json
NyCodeGHG: I used IntelliJ's rename feature here hoping it would rename everything, but your suggestion is probably way better with a more expressive description.

app/Http/Controllers/Frontend/Transport/StatusController.php
NyCodeGHG: I tried it and it fired on a destination change, but yeah this should be a test case

resources/lang/de.json
jeyemwey: Fixed with 72936a54386226f5a468be1ec97126e046ee1547.

Thaaankssss

app/Models/User.php
NyCodeGHG: Webhooks are not in the response for that endpoint

NyCodeGHG:
the settings ui for webhooks now looks like this:
screenshot

I thought the explanation is a good idea because i don't think most users know what a webhook is. But I think this wall of text is a bit too much and maybe we should put it behind a Questionmark Symbol or something similar?
Feedback is greatly appreciated

NyCodeGHG:
So @DRSchlaubi and I had a few thoughts about the flow of the creation of a webhook. (See https://github.com/Traewelldroid/traewelldroid/issues/65)

The current flow would be that an application can just create a webhook when it authorized with the webhook scope.
That way is very intransparent to the user and applications could theoretically create hundreds of webhooks if we don't limit them.
@DRSchlaubi had an idea to introduce an OAuth-like flow for the user to authorize the creation of a webhook and the respective webhook events the app will get access to. With that every webhook would have to get authorized by the user and he would explicitly know what the third party app gets notified about.

DRSchlaubi:
IMO D[iscord has a pretty good implementation of webhooks](https://discord.com/developers/docs/topics/oauth2#webhooks) that we could adapt that.

However, we maybe can add some slight modifications to it to make it easier for mobile clients (Like traewelldroid) to implement push notifications
- Apps can specify that they do support notifications (or something else needing webhooks)
  - Then that "webhook permission" can be granted either right on the first authorization (perhaps using an optional checkbox whilst authorizing)
  - Or later using an OAuth flow like Discord's
- Somehow mobile clients will need to know whether that webhooks exist
  - Either through the existing API, by looking up all webhooks and searching for a specific name
    - which would be tedious and require the webhook scope just to know whether the webhook is set up or not
   - Or through a different endpoint, that can validate whether that dedicated "notifications webhook" exist or not

I have no idea how far this is doable using passport, but I wanted to offer my ideas

jeyemwey:
> I have no idea how far this is doable using passport

FWIW nobody at Trwl really _understands_ API scopes in Passport/Sanctum, however I would cross that bridge when we come to it (namely once we implement API scopes).

> The current flow would be that an application can just create a webhook when it authorized with the webhook scope.
That way is very intransparent to the user and applications could theoretically create hundreds of webhooks if we don't limit them.

I thought of pushing a `Application X requested to receive webhooks for status edits` notification to the user via the given notification system. Maybe, we can even add an `active` flag so the user has to confirm that they're okay with each registered webhook?

> Somehow mobile clients will need to know whether that webhooks exist

NyCodeGHG has proposed a `GET /webhooks` endpoint which filters the webhooks by user and the requesting application. That looks sufficient to me at the moment:

https://github.com/Traewelling/traewelling/blob/0d4a658a3563b9a8ac2592509af447fd0ec122b2/app/Http/Controllers/API/v1/WebhookController.php#L16-L22

NyCodeGHG:
What I've planned for the creation flow of a webhook right now is the following:
* App initiates usual OAuth Flow on /oauth/authorize?client_id=blabla
  * We extend that flow with the optional `trwl_webhook_events` and `trwl_webhook_url` parameters
  * With this approach a user always has to approve the creation of a webhook and they know exactly which app gets notified about stuff
  * When we modify this flow we can also show this in the Authorization UI and when we're on it, we can improve the page to match traewellings style
* When the user approves the authorization, the resulting OAuth response will contain a webhook field containing the id and secret of the webhook similar to [how discord is doing that](https://discord.com/developers/docs/topics/oauth2#webhooks)

NyCodeGHG:
Here is a screenshot of the current state of the OAuth screen
grafik
We should maybe fix the behaviour of the footer so it's on the bottom of the page when the content is not big enough

DRSchlaubi:
> * We extend that flow with the optional `trwl_webhook_events` and `trwl_webhook_url` parameters
@NyCodeGHG 

And those would be usable both with other scopes (so you can request an access token at the same time) and without (so you can request just a webhook), correct?

> NyCodeGHG has proposed a GET /webhooks endpoint which filters the webhooks by user and the requesting application. That looks sufficient to me at the moment:

But ig that will require a specific scope, but ig if the user didn't grant the application a webhook scope that application won't have any webhook setup, so it would be fine

NyCodeGHG:
> And those would be usable both with other scopes (so you can request an access token at the same time) and without (so you can request just a webhook), correct?

yes

> But ig that will require a specific scope, but ig if the user didn't grant the application a webhook scope that application won't have any webhook setup, so it would be fine

i don't think here is a scope necessary because an application can only retrieve its own webhooks.
We could add a "webhook management token" to the response to allow the application to delete the webhook if needed.

DRSchlaubi:
Looks like we're good to go

app/Providers/AuthServiceProvider.php
```diff
@@ -29,7 +30,14 @@ class AuthServiceProvider extends ServiceProvider
...
-    public function boot(): void {
+    public function boot(): void
+    {
         $this->registerPolicies();
+
+        Route::group(['prefix' => 'oauth', 'as' => 'oauth.'], function () {
+            Route::get('authorize', [AuthorizationController::class, 'authorize'])
+                ->middleware(['web'])
+                ->name('authorizations.authorize');
+        });
```
MrKrisKrisu: Can you move this to the `routes/web.php`-file or is there a reason why the route is defined here?

app/Providers/EventServiceProvider.php
```diff
@@ -34,9 +48,8 @@ public function boot(): void {
         parent::boot();
 
         // Dispatch Jobs from Events
-        Event::listen(fn(UserCheckedIn $event)
-            => PostStatusOnTwitter::dispatchIf($event->shouldPostOnTwitter, $event->status));
-        Event::listen(fn(UserCheckedIn $event)
-            => PostStatusOnMastodon::dispatchIf($event->shouldPostOnMastodon, $event->status, $event->shouldChain));
+        Event::listen(fn(UserCheckedIn $event) => PostStatusOnTwitter::dispatchIf($event->shouldPostOnTwitter, $event->status));
+        Event::listen(fn(UserCheckedIn $event) => PostStatusOnMastodon::dispatchIf($event->shouldPostOnMastodon, $event->status, $event->shouldChain));
+        Event::listen(fn(WebhookCallFailedEvent $event) => Log::error("Webhook call failed: " . json_encode($event)));
```
MrKrisKrisu: Maybe put this in a context?

```suggestion
        Event::listen(fn(WebhookCallFailedEvent $event) => Log::error("Webhook call failed.", ['event' => json_encode($event)]));
```

routes/api.php
```diff
@@ -1,4 +1,4 @@
-<?php
+<? php
```
MrKrisKrisu: Not sure about that. :D 

image

app/Providers/AuthServiceProvider.php
NyCodeGHG: They need to be registered there, when they are defined in `routes/web.php` they do not correctly override the passport routes.

MrKrisKrisu: Oh, okay! Good 2 know.

routes/api.php
NyCodeGHG: oops, i think this is my formatter's fault

app/Listeners/NotificationSentWebhookListener.php
```diff
@@ -0,0 +1,20 @@
...
+use App\Http\Controllers\Backend\WebhookController;
+use App\Models\User;
+use Illuminate\Notifications\DatabaseNotification;
+use Illuminate\Notifications\DatabaseNotificationCollection;
+use Illuminate\Notifications\Events\NotificationSent;
+use Illuminate\Support\Facades\Log;
+
+class NotificationSentWebhookListener
+{
+    public function handle(NotificationSent $event)
```
jeyemwey, line 11: We probably need a check here for the type of Notification that is sent since Laravel uses the system for all sorts of stuff  (such as the verification email and password reset emails).

IMO, ˋ$event->channel === 'database'` should work to filter out unwanted notifications.

NyCodeGHG:
I have some more ideas to improve the experience for third party developers, but I think it's better to do that in additional PRs, this one is already huge enough. If I didn't forget something important I think this would be ready for review/merge.

NyCodeGHG:
There is also a need for a place to document the behaviour of the customized OAuth flow, but I'm unsure whats the best place to put that? Maybe an external site built using [Astro](https://astro.build) or [Docusaurus](https://docusaurus.io/) would be better for that, as in "Developer Documentation" for träwelling.

app/Enum/WebhookEvent.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
```
MrKrisKrisu, line 3: Can you add that to all new files you create? :)

app/Http/Controllers/Backend/WebhookController.php
```diff
@@ -0,0 +1,151 @@
...
+     * @param Client|null $client
+     * @param int         $webhookId
+     *
+     * @return bool
+     * @throws PermissionException
+     */
+    public static function deleteWebhook(
+        User   $user,
+        Client|null $client,
+        int    $webhookId
```
MrKrisKrisu: Can you change this function to receive directly an `Webhook $webhook` object instead of the id?

app/Http/Controllers/Backend/WebhookController.php
```diff
@@ -0,0 +1,151 @@
...
+    ): bool {
+        $webhook = Webhook::find($webhookId);
+
+        if ($webhook === null) {
+            throw new ModelNotFoundException();
+        }
+        if ($client != null) {
+            if ($user->id != $webhook->user->id || $client->id != $webhook->oauthClient->id) {
```
MrKrisKrisu: Can you add an [policy](https://laravel.com/docs/9.x/authorization#writing-policies) for that?

NyCodeGHG:
We might want to restrict webhook urls to https only in production i think? Thats missing right now

app/Http/Controllers/Backend/WebhookController.php
```diff
@@ -0,0 +1,151 @@
...
+                ->withHeaders([
+                    'X-Trwl-User-Id'    => $user->id,
+                    'X-Trwl-Webhook-Id' => $webhook->id,
+                ])
+                ->payload([
+                    'event' => $event->name(),
+                    ...$data
+                ])
+                ->useSecret($webhook->secret)
+                ->dispatch();
```
MrKrisKrisu, line 103: I'm not sure if we maybe want to expect a specific response from the other server to make sure we don't spam "into the void".

Maybe just a blank page with an "OK" or a specific HTTP code? Should be discussed. What are other opinions?

app/Http/Controllers/Backend/WebhookController.php
```diff
@@ -0,0 +1,151 @@
...
+     */
+    public static function createWebhookRequest(
+        User $user,
+        Client $client,
+        string $oauth_code,
+        string $url,
+        int  $events,
+    ): WebhookCreationRequest {
+        return WebhookCreationRequest::create([
+            'id' => hash('sha256', $oauth_code),
```
MrKrisKrisu:
```suggestion
        string $oauthCode,
        string $url,
        int  $events,
    ): WebhookCreationRequest {
        return WebhookCreationRequest::create([
            'id' => hash('sha256', $oauthCode),
```

app/Http/Controllers/Backend/WebhookController.php
```diff
@@ -0,0 +1,151 @@
...
+            'expires_at' => Carbon::now()->addHour(),
+            'events' => $events,
+            'url' => $url,
+        ]);
+    }
+
+    public static function findWebhookRequest(
+        string $oauth_code,
+    ): WebhookCreationRequest | null {
+        return WebhookCreationRequest::where('id', hash('sha256', $oauth_code))->first();
```
MrKrisKrisu:
```suggestion
        string $oauthCode,
    ): WebhookCreationRequest|null {
        return WebhookCreationRequest::where('id', hash('sha256', $oauthCode))->first();
```

app/Http/Controllers/Backend/WebhookController.php
```diff
@@ -0,0 +1,151 @@
...
+                ->payload([
+                    'event' => $event->name(),
+                    ...$data
+                ])
+                ->useSecret($webhook->secret)
+                ->dispatch();
+        }
+    }
+
+    public static function deleteAllWebhooks(User $user, Client $client)
```
MrKrisKrisu:
```suggestion
    public static function deleteAllWebhooks(User $user, Client $client): void
```

app/Http/Controllers/Backend/WebhookController.php
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public static function sendNotificationWebhook(User $user, DatabaseNotification $notification)
+    {
+        self::dispatchWebhook($user, WebhookEvent::NOTIFICATION, [
+            'notification' => new UserNotificationResource($notification)
+        ]);
+    }
+
+    static function dispatchWebhook(User $user, WebhookEvent $event, array $data)
```
MrKrisKrisu:
```suggestion
    static function dispatchWebhook(User $user, WebhookEvent $event, array $data): void
```

app/Http/Controllers/Backend/WebhookController.php
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public static function sendStatusWebhook(Status $status, WebhookEvent $event)
+    {
+        self::dispatchWebhook($status->user, $event, [
+            'status' => new StatusResource($status)
+        ]);
+    }
+
+    public static function sendNotificationWebhook(User $user, DatabaseNotification $notification)
```
MrKrisKrisu:
```suggestion
    public static function sendNotificationWebhook(User $user, DatabaseNotification $notification): void
```

app/Http/Controllers/Backend/WebhookController.php
```diff
@@ -0,0 +1,151 @@
...
+        if ($client != null) {
+            if ($user->id != $webhook->user->id || $client->id != $webhook->oauthClient->id) {
+                throw new PermissionException();
+            }
+        }
+        $webhook->delete();
+        return true;
+    }
+
+    public static function sendStatusWebhook(Status $status, WebhookEvent $event)
```
MrKrisKrisu:
```suggestion
    public static function sendStatusWebhook(Status $status, WebhookEvent $event): void
```

app/Http/Controllers/Frontend/DevController.php
```diff
@@ -3,14 +3,16 @@
...
+use App\Repositories\OAuthClientRepository;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
 use Illuminate\View\View;
 use Laravel\Passport\ClientRepository;
 
 class DevController extends Controller
 {
-    public function renderAppList(): View {
+    public function renderAppList(): View
```
MrKrisKrisu: Hmm, currently we have the braces on the same line. :D Can you change your formatter for this project to match this? <3

app/Http/Controllers/Frontend/DevController.php
```diff
@@ -33,50 +36,70 @@ public function renderUpdateApp(int $appId): View {
...
+    public function updateApp(int $appId, Request $request): RedirectResponse
+    {
         $validated = $request->validate([
-                                            'name'     => ['required', 'string'],
-                                            'redirect' => ['required', 'string'],
-                                        ]);
+            'name'                   => ['required', 'string'],
+            'redirect'               => ['required', 'string'],
+            'enable_webhooks'        => ['nullable'],
+            'authorized_webhook_url' => ['nullable', 'url'],
```
MrKrisKrisu:
```suggestion
            'authorized_webhook_url' => ['nullable', 'url', 'starts_with:https'],
```

Maybe this is one option to force https. (untested)

app/Http/Controllers/Frontend/DevController.php
```diff
@@ -33,50 +36,70 @@ public function renderUpdateApp(int $appId): View {
...
         $validated = $request->validate([
-                                            'name'         => ['required', 'string'],
-                                            'redirect'     => ['required', 'string'],
-                                            'confidential' => ['nullable'],
-                                        ]);
+            'name'                   => ['required', 'string'],
+            'redirect'               => ['required', 'string'],
+            'confidential'           => ['nullable'],
+            'enable_webhooks'        => ['nullable'],
+            'authorized_webhook_url' => ['nullable', 'url'],
```
MrKrisKrisu:
```suggestion
            'authorized_webhook_url' => ['nullable', 'url', 'starts_with:https'],
```

database/migrations/2023_01_16_033114_create_webhooks_table.php
```diff
@@ -0,0 +1,42 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up()
+    {
+        Schema::create('webhooks', function (Blueprint $table) {
+            $table->id();
+            $table->timestamps();
```
MrKrisKrisu: Can you put `->timestamps()` at the end, so `created_at` and `updated_at` are the last columns?

database/migrations/2023_01_16_033114_create_webhooks_table.php
```diff
@@ -0,0 +1,42 @@
...
+                ->cascadeOnDelete();
+            $table->foreignId('user_id')
+                ->constrained()
+                ->cascadeOnUpdate()
+                ->cascadeOnDelete();
+            $table->text('url');
+            $table->text('secret')->nullable();
+            $table->unsignedInteger('events');
+        });
+        Schema::create('webhook_creation_requests', function (Blueprint $table) {
```
MrKrisKrisu: Can you put this in an additional migration?

database/migrations/2023_02_04_184426_extend_oauth_clients_with_custom_data.php
```diff
@@ -0,0 +1,36 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('oauth_clients', function (Blueprint $table) {
+            $table->string('privacy_policy_url')->nullable();
+            $table->boolean('webhooks_enabled')->default(false);
+            $table->string('authorized_webhook_url')->nullable();
```
MrKrisKrisu:
```suggestion
            $table->string('authorized_webhook_url')->nullable()->after('redirect');
            $table->string('privacy_policy_url')->nullable()->after('redirect');
            $table->boolean('webhooks_enabled')->default(false)->after('redirect');
```

Just reordering, that the columns are created after `redirect` and in the order from bottom to top. (yes, bottom to top... ._.)

MrKrisKrisu (changes requested):
First of all: Thank you very much for your work! This is an great PR! <3

I haven't looked at everything in detail yet and will also test the PR locally soon. But I already have a few minor comments that I noticed while scrolling through.

MrKrisKrisu:
@HerrLevin Can you have a look at the API Docs?

app/Http/Controllers/Backend/WebhookController.php
jeyemwey: I don't know if we even have a retry mechanism, if so I'd check if the response was in 200-299. This seems to be Common practice among the Webhook APIs, see i.e. [stripe](https://stripe.com/docs/webhooks/best-practices).

NyCodeGHG: [laravel-webhook-server](https://github.com/spatie/laravel-webhook-server) already handles response code checking and retries.

jeyemwey: Ack, then I'd keep it as is. Thanks for providing the link.

app/Http/Controllers/Frontend/DevController.php
NyCodeGHG: sure, i'll do that. Besides the braces, the formatting will mismatch i think, because, i assume most if not all of you use phpstorm and it has its own formatter?

app/Enum/WebhookEvent.php
NyCodeGHG: to all other files i created? I'm not sure if thats what you meant, because right now `declare(strict_types=1);` is only used in the enum files

MrKrisKrisu: Yes, this would be great. But it is non blocking for me.
Strict types are nice 2 have at the beginning in a file, so we are forced to use explizit types in every file and improve our type safety.

app/Http/Controllers/Backend/WebhookController.php
NyCodeGHG: I moved the check into a policy, but I found no way for moving the client id equality check into the policy. Do you have an idea how that could be done? Or should I leave it as it is right now?

app/Http/Controllers/Frontend/DevController.php
NyCodeGHG: i think making this configurable would be better, to allow easier local testing using an http proxy

app/Enum/WebhookEvent.php
NyCodeGHG: i think it would be better to do this in an extra PR for the whole project when this PR is merged

README.md
```diff
@@ -48,7 +48,7 @@ To set up a Träwelling instance you'll need:
 
 * [npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
 * [MariaDB](https://mariadb.org/download) or [MySQL](https://www.mysql.com/de/downloads/) (SQLite is used for running
-  tests)
+  tests) (You can use [Docker Compose](https://docs.docker.com/compose/) and the [dev.docker-compose.yml](dev.docker-compose.yml) for that)
```
jeyemwey, line 51: Why is there a need for _another_ docker-compose file?

tests/TestCase.php
```diff
@@ -169,6 +171,31 @@ protected function createAdminUser(): User {
         return $admin;
     }
 
+    public function createClient(User $user): PassportClient {
```
jeyemwey: Can we rename this to `createWebhookClient` please?

app/Http/Controllers/API/v1/WebhookController.php
```diff
@@ -0,0 +1,140 @@
...
+     *     )
+     *
+     * @param int $webhookId
+     *
+     * @return JsonResponse
+     */
+    public function deleteWebhook(Request $request, int $webhookId): JsonResponse {
+        try {
+            $webhook = Webhook::findOrFail($webhookId);
+            $this->authorize('delete', $webhook);
```
jeyemwey: Can we move this to the backend controller as well? Then we have all the reasons why a 403 might throw together.

app/Http/Controllers/Backend/Auth/AccessTokenController.php
```diff
@@ -0,0 +1,53 @@
...
+    public function issueToken(ServerRequestInterface $requestInterface) {
+        return $this->withErrorHandling(function () use ($requestInterface) {
+            return $this->extendResponseWithWebhookData(
+                $requestInterface,
+                $this->server->respondToAccessTokenRequest($requestInterface, new Psr7Response)
+            );
+        });
+    }
+
+    function extendResponseWithWebhookData(ServerRequestInterface $requestInterface, Psr7Response $response) {
```
jeyemwey:
```suggestion
    public function extendResponseWithWebhookData(ServerRequestInterface $requestInterface, Psr7Response $response): Psr7Response {
```

routes/web.php
```diff
@@ -11,6 +11,7 @@
 |
 */
 
+use App\Http\Controllers\Backend\Auth\AuthorizationController;
```
jeyemwey:
```suggestion
```

unused

routes/web.php
```diff
@@ -37,44 +39,51 @@
 use App\Http\Controllers\UserController;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Route;
+use Laravel\Passport\Http\Controllers\AccessTokenController;
+use Laravel\Passport\Http\Controllers\ApproveAuthorizationController;
+use Laravel\Passport\Http\Controllers\AuthorizedAccessTokenController;
+use Laravel\Passport\Http\Controllers\ClientController;
+use Laravel\Passport\Http\Controllers\DenyAuthorizationController;
+use Laravel\Passport\Http\Controllers\PersonalAccessTokenController;
+use Laravel\Passport\Http\Controllers\ScopeController;
```
jeyemwey:
```suggestion
```

unused

dev.docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+# Docker Compose environment for local development
```
jeyemwey, line 1: I dont understand why we need another docker-compose. (see above)

app/Http/Controllers/Backend/Auth/ApproveAuthorizationController.php
```diff
@@ -0,0 +1,36 @@
...
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Log;
+use Laravel\Passport\Http\Controllers\ApproveAuthorizationController as PassportApproveAuthorizationController;
+
+class ApproveAuthorizationController extends PassportApproveAuthorizationController {
+    public function approve(Request $request) {
+        $webhook = $request->session()->get('webhook');
+        $response = parent::approve($request);
+        if ($webhook) {
+            parse_str(parse_url($response->headers->get("Location"))["query"], $query);
```
jeyemwey, line 16: whats the reason behind this line? It isn't put into any variable or anything?

app/Http/Controllers/Backend/WebhookController.php
```diff
@@ -0,0 +1,133 @@
...
+use Laravel\Passport\Client;
+use Spatie\WebhookServer\WebhookCall;
+
+abstract class WebhookController extends Controller {
+    public static function index(User $user): object {
+        return $user->webhooks;
+    }
+
+    public static function createWebhook(
+        WebhookCreationRequest $request
```
jeyemwey, line 29: Can the WebhookCreationRequest be deleted afterwards? Just to not pollute the database more than necessary?

app/Events/StatusUpdateEvent.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace App\Events;
+
+use App\Models\Status;
+use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Facades\Log;
+
+class StatusUpdateEvent {
```
jeyemwey, line 10: This also needs to be dispatched from `\App\Http\Controllers\Frontend\Admin\StatusEditController::edit`

app/Providers/AuthServiceProvider.php
```diff
@@ -31,5 +38,20 @@ class AuthServiceProvider extends ServiceProvider
      */
     public function boot(): void {
         $this->registerPolicies();
+
+        Passport::useClientModel(OAuthClient::class);
+
+        // Override passport routes
+        Route::group(['prefix' => 'oauth', 'as' => 'oauth.'], function () {
```
jeyemwey, line 73: Can we move this to `routes/web.php`? Just so we have everything in one place.

resources/views/settings/cards/webhooks.blade.php
```diff
@@ -0,0 +1,47 @@
...
+                        <tr>
+                            <td>{{ $webhook->client->name }}</td>
+                            <td>{{ Carbon::parse($webhook->created_at)->isoFormat(__('datetime-format')) }}</td>
+                            <td>
+                                <ul>
+                                    @foreach($webhook->events as $event)
+                                        <li>{{ __('settings.webhook_event.' . $event)}}</li>
+                                    @endforeach
+                                </ul>
+                            </td>
```
jeyemwey: For some reasons, this doesn't work for me.

image

routes/web.php
```diff
@@ -158,106 +167,108 @@
...
+            ->name('settings.delete-profile-picture');
 
         Route::post('/delsession', [UserController::class, 'deleteSession'])
-             ->name('delsession');
+            ->name('delsession');
         Route::post('/deltoken', [UserController::class, 'deleteToken'])
-             ->name('deltoken');
+            ->name('deltoken');
+        Route::post('/delwebhook', [WebhookController::class, 'deleteWebhook'])
+            ->name('delwebhook');
```
jeyemwey: Are all the changes in this file, apart from the `delwebhook` route necessary? Would like to clean up..

tests/TestCase.php
```diff
@@ -13,20 +13,21 @@
...
 use Illuminate\Foundation\Testing\TestCase as BaseTestCase;
 use Illuminate\Support\Collection;
 use Illuminate\Testing\TestResponse;
 use JetBrains\PhpStorm\ArrayShape;
+use Laravel\Passport\Client as PassportClient;
 use Tests\Feature\CheckinTest;
 
-abstract class TestCase extends BaseTestCase
-{
+abstract class TestCase extends BaseTestCase {
```
jeyemwey, line 30: Can you create a test which tests the creation of an oauth client and token, that also checks if the webhook is created successfully? That would allow us to have a documented way how the flow should work.

tests/Feature/Webhooks/WebhookStatusTest.php
```diff
@@ -0,0 +1,216 @@
...
+    public function testWebhookSendingOnLike() {
+        Bus::fake();
+
+        $user = $this->createGDPRAckedUser();
+        $client = $this->createClient($user);
+        $this->createWebhook($user, $client, [WebhookEvent::CHECKIN_UPDATE]);
+        $status = $this->createStatus($user);
+        StatusController::createLike($user, $status);
+
+        Bus::assertDispatched(function (CallWebhookJob $job) use ($status) {
```
jeyemwey, line 78:
```suggestion
        // For self-likes, a CHECKIN_UPDATE is sent, but no notification.
        Bus::assertDispatched(function (CallWebhookJob $job) use ($status) {
```

jeyemwey (commented):
One more bunch of questions and feedback, but after that I feel confident that we can finally merge it. Thank you again for taking on this massive work.

README.md
NyCodeGHG: The other docker-compose files (don't ask me why there are 2) seem to be for running in production. The dev.docker-compose.yml (as the name suggests) is for running a local database for local testing.

app/Providers/AuthServiceProvider.php
NyCodeGHG: This has already been discussed [here](https://github.com/Traewelling/traewelling/pull/1304#discussion_r1089816858)
This has to be here, otherwise the routes do not get overwritten

routes/web.php
NyCodeGHG: No, that was my formatter going crazy, sorry for that

README.md
jeyemwey: Okay thanks

app/Http/Controllers/Backend/Auth/ApproveAuthorizationController.php
NyCodeGHG: [parse_str](https://www.php.net/manual/en/function.parse-str.php) returns `void`. The second parameter is a result type where the result gets put into.
Idk why it's like this, it's very confusing

NyCodeGHG:
I don't know why the previous checks run failed, because that was a test I didn't touch and now it passed?
Seems like we have a flaky test there

jeyemwey:
> I don't know why the previous checks run failed, because that was a test I didn't touch and now it passed?
> Seems like we have a flaky test there

That just happens sometimes. Looks like it worked on the next try.

tests/TestCase.php
NyCodeGHG: The OAuth flow is really complicated to test and I think it would be way easier (and easier to maintain) to do this in an actual browser e2e test rather than in an integration test.
I have not really found a way to do this in an integration test without touching passport internals.

jeyemwey: Okay no worries, we can add that later as well.

NyCodeGHG:
I just realized that there's no way of deleting a webhook from the application side when the user's token has been revoked

app/Enum/WebhookEvent.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace App\Enum;
+
+enum WebhookEvent: int {
+    case CHECKIN_CREATE = 1 << 0;
```
HerrLevin, line 8: Why exactly do you use bit shift here? Not really a problem for me. I'm just wondering.

app/Http/Controllers/Backend/Auth/ApproveAuthorizationController.php
HerrLevin, line 16: (Note to self: Maybe we can change this to use laravel methods)

app/Http/Controllers/Backend/Auth/AuthorizationController.php
```diff
@@ -0,0 +1,141 @@
...
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Laravel\Passport\ClientRepository  $clients
+     * @param  \Laravel\Passport\TokenRepository  $tokens
+     * @return \Illuminate\Http\Response
+     */
+    public function authorize(
+        ServerRequestInterface $psrRequest,
+        Request $request,
+        ClientRepository $_,
+        TokenRepository $tokens
```
HerrLevin, line 36: Note to self: can we reduce this code with parent::authorize, to maybe reduce future complications, once passport get's a major update?

resources/lang/de.json
```diff
@@ -618,6 +618,25 @@
...
+    "settings.webhook-event-notifications-description": "Aktivitäten",
+    "settings.webhook_event.checkin_create": "Erstellen eines Checkins",
+    "settings.webhook_event.checkin_update": "Bearbeiten eines Checkins",
+    "settings.webhook_event.checkin_delete": "Löschen eines Checkins",
+    "settings.webhook_event.notification": "Erhalten einer Benachrichtigung",
+    "menu.oauth_authorize.title": "Autorisierung",
+    "menu.oauth_authorize.authorize": "Autorisieren",
+    "menu.oauth_authorize.cancel": "Abbrechen",
+    "menu.oauth_authorize.webhook_request": "Diese Anwendung möchte über die folgenden Aktivitäten auf deinem Träwelling Account benachrichtigt werden:",
+    "menu.oauth_authorize.request": "<span class=\"fw-bold\">:application</span> is requesting permission to access your account.",
```
HerrLevin, line 634: This is not German. :D

HerrLevin (approved):

app/Enum/WebhookEvent.php
NyCodeGHG: I tried using a regular enum, but that was kinda annoying, this worked out better for me and is probably also more efficient

PR #1312, opened 2023-01-21
:children_crossing: Improve ui for connected services

MrKrisKrisu:
image


HerrLevin:
I'd actually move these new connected services to API-Tokens. This menu implies that WE get data from these services. Not that these services can acquire data from the authenticated user.

MrKrisKrisu:
You're right.

PR #1313, opened 2023-01-22
:sparkles: Add button to "join" other check ins

DRSchlaubi:
This PR adds a pretty popular [Träwelldroid](https://github.com/traewelldroid) feature to Träwelling itself, a button to join other peoples checkins with a single click.

image

Curretnly there is an issue I don't know how to fix,: If you delete your checkin after joining you cannot join again because of an AlreadyCheckedInException


jeyemwey:
Hi, thank you sooo much for your contribution! There are a few things that I'd need to have changed so your change works reliably and efficiently:

1. The `includes/check-in-modal.blade.php is more or less an exact copy of the modal in `trips.blade.php`. It would be great to have that deduplicated and imported, so that we can't "miss" changes in either version of the same form.
2. Since the `includes/status.blade` is used at _so many_ places (i.e. activeJourneys, single view, profiles, event view), I'd suggest moving the `EventBackend::activeEvents()` request inward into the include (or the check-in-modal for that matter).

> Currently, there is an issue I don't know how to fix: If you delete your checkin after joining you cannot join again because of an AlreadyCheckedInException

Uhm, that doesn't sound great. I can't really see how the delete is changed when there's a cloned checkin (Maybe @MrKrisKrisu can help?)

MrKrisKrisu:
> Maybe @MrKrisKrisu can help?

Unfortunately, I currently have no capacity. I try to look at the PR at the weekend.

DRSchlaubi:
> The includes/check-in-modal.blade.php is more or less an exact copy of the modal in trips.blade.php`. It would be great to have that deduplicated and imported, so that we can't "miss" changes in either version of the same form.

That was actually the entire Point of this, but lookslike i forgot to die that 😂

> Since the includes/status.blade is used at so many places (i.e. activeJourneys, single view, profiles, event view), I'd suggest moving the EventBackend::activeEvents() request inward into the include (or the check-in-modal for that matter).

I thought whether there ist a better way to so this, but i don't know PHP good enough to know how to do this

resources/views/includes/check-in-modal.blade.php
```diff
@@ -0,0 +1,106 @@
...
+                                                </span>
+                                        </label>
+                                    </div>
+                                @endif
+                            @endif
+                        @endif
+                        @include('includes.business-dropdown')
+                        @include('includes.visibility-dropdown')
+                    </div>
+
```
jeyemwey, line 67: > I thought whether there ist a better way to so this, but i don't know PHP good enough to know how to do this

You can just set `$events` it in the view itself:

```suggestion

@php($events = \App\Http\Controllers\Backend\EventController::activeEvents())
```

DRSchlaubi: Oh so that's how this works thx

MrKrisKrisu (approved):

PR #1319, opened 2023-01-23
Fix shields.io issues

DRSchlaubi:
See https://github.com/badges/shields/issues/8671

I have no idea which file the Resources with dev one is, as I cannot find it


jeyemwey:
Thanks for noting :))

PR #1331, opened 2023-01-28
:fire: Cleanup notification classes

MrKrisKrisu:


MrKrisKrisu:
@jeyemwey I don't see any reason for the `via`-functions. Can you confirm me, that I can remove these?
Notifications work for me without.

jeyemwey:
> I don't see any reason for the `via`-functions. Can you confirm me, that I can remove these? Notifications work for me without.

`via` is [part of the Laravel Notification Interface](https://laravel.com/docs/9.x/notifications#specifying-delivery-channels). I'm not sure how deleting the method might impact the delivery/saving of notifications. If you don't want to specify it every time, I'd suggest moving it to the BaseNotification class.

MrKrisKrisu:
We don't have other broadcast channels registered, but yeah: We'll go the safe way and I added the function to the BaseController. :)

jeyemwey:
> We don't have other broadcast channels registered

Yes we do, EmailVerification and PasswordReset emails are sent via a notification.

MrKrisKrisu:
not important i guess. :D

PR #1335, opened 2023-01-29
:bug: Fix check if arrival is really cancelled

MrKrisKrisu:


jeyemwey:
Still missing here:

- [x] `$lastStopover` shortcut in `FrontendTransportController`
- [ ] Amend `Virtual/Models/TrainStopover.php` with the new properties

MrKrisKrisu:
> $lastStopover shortcut in FrontendTransportController

Well, our fast check-in is a feature which isn't in my mind. Thanks!

> Amend Virtual/Models/TrainStopover.php with the new properties

But why? This attribute isn't added to the api resources, so it's not required in the virtual model?

jeyemwey:
> But why? This attribute isn't added to the api resources, so it's not required in the virtual model?

Ack, thanks for clarifying. LGTM :)

jeyemwey (approved):

PR #1344, opened 2023-01-30
✨💥 Add scopes to API 

HerrLevin:
⚠️ This PR adds scopes to the API. We'll add __all__ scopes to existing API tokens to reduce need for user migration. We're also creating all new Bearer Tokens with all scopes, since there's no way to add these in the old auth flow. 

All OAuth Apps will need to migrate their auth flow to use scopes from now on. (See API documentation)


MrKrisKrisu:
> We'll add **all** scopes to existing API tokens to reduce need for user migration. 

Please write a migration for it and don't do it manually, if possible.

> We're also creating all new Bearer Tokens with all scopes, since there's no way to add these in the old auth flow.

Here also please.

> All OAuth Apps will need to migrate their auth flow to use scopes from now on. (See API documentation)

So a lot of apps won't work after this PR? We should announce this bigger beforehand and not release it so abruptly.

jheubuch:
Can we have consistent scope-names?
Like `operation-resource` like `read-statuses` or `write-likes`
At the moment there is e.g. `settings-read`, which destroys this consistency

app/Providers/AuthServiceProvider.php
```diff
@@ -24,12 +24,49 @@ class AuthServiceProvider extends ServiceProvider
...
+        'write-event-suggestions'        => 'suggest events in your name',
+        'write-support-tickets'          => 'create support tickets in your name',
+        'settings-read'                  => 'see your settings, email, etc.',
+        'settings-write-profile'         => 'edit your profile',
+        'settings-read-profile'          => 'see your profile data, e.g. email',
+        'settings-write-mail'            => 'edit your email',
+        'settings-write-profile-picture' => 'edit your profile picture',
+        'settings-write-privacy'         => 'change your privacy settings',
+        'settings-read-followers'        => 'show follow-requests and followers',
+        'settings-write-calendar'        => 'create and delete new calendar-tokens',
```
MrKrisKrisu: see https://github.com/Traewelling/traewelling/pull/1344#issuecomment-1408961909

MrKrisKrisu (changes requested):

database/migrations/2023_01_30_173822_add_scopes_to_existing_oauth_access_tokens.php
```diff
@@ -0,0 +1,16 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void {
+        Schema::table('oauth_access_tokens', function(Blueprint $table) {
+            DB::table('oauth_access_tokens')->update(['scopes' => '["' . implode('","', array_keys(AuthServiceProvider::$scopes)) . '"]']);
```
MrKrisKrisu, line 15: I would not give permission to access **_everything_** by default. Would exclude the `extra` scopes? For future keys, the user must and can then decide for himself.

MrKrisKrisu (changes requested):

HerrLevin:
> `settings-read`, which destroys this consistency

I wanted to group settings in their own category because in my head these are scopes of higher importance. If that idea is only reasonable in my head, then i'll change that. (cc @MrKrisKrisu @jheubuch)

MrKrisKrisu:
> I wanted to group settings in their own category because in my head these are scopes of higher importance. If that idea is only reasonable in my head, then i'll change that. (cc @MrKrisKrisu @jheubuch)

I would agree to @jheubuch and think a consistent scope-names is more important then "mark" scopes with higher permission requirements. In addition the `extra-` scopes are okay for me in this format.

PR #1345, opened 2023-01-30
DRAFT: Improve vehicle sequence support 

MrKrisKrisu:


MrKrisKrisu:
This PR will remain open as a memory for a while. For the time being, I will completely remove the carriage sequences support from Trwl. The feature doesn't currently work anyway due to API changes. The feature should be redeveloped and developed properly.

PR #1371, opened 2023-02-06
:children_crossing: Remove dropdown animations

vainamov:
During the fadeout animation, the dropdown items still receive click events and block interaction with the underlying elements. When for example a user tries to check-in after changing the type to `commute`, often the animation hasn't finished yet which results in the dropdown item for `business` reacting to the click event instead of the check-in button.

Other options to solve this problem could be using CSS to set `pointer-events: none` and reverting it to `auto` when the dropdown is open or perhaps increasing the animation speed. But I felt removing the animation makes the UI feel snappier.


MrKrisKrisu (approved):
Omg, thanks for this! This is extremely annoying to me as well. I have too fast fingers.

PR #1374, opened 2023-02-06
:lipstick: Improve styling of the search autocomplete buttons

### LINKED ISSUE #1283: Hover effect on "Last station" buttons is broken
**Describe the bug**
- When the two buttons next to the "last station" input on the start page are hovered, the border disappears (tested in latest Chrome and Firefox on macOS)
- Icons inside these buttons are not centered correctly

**Expected behavior**
Border should not disappear on hover.

**Screenshots**
Screenshot 2023-01-07 at 11 42 04

vainamov:
Fixes #1283


MrKrisKrisu:
image
image

The hover effect above the left button is great! Thank you.
Can you do the same with the right button? The left bar is still missing there.

vainamov:
By design, MDB buttons in input groups have no left border to create the attached look. MDB doesn't change the border-color on hover (on every outlined button actually) because of this exact reason and that's the only solution I can think of as well.

I would set the border-color for `.stationSearchButton:hover` and `:focus` to the same `#bdbdbd` as it is without interaction, if you're okay with that.

MrKrisKrisu:
> By design, MDB buttons in input groups have no left border to create the attached look. MDB doesn't change the border-color on hover (on every outlined button actually) because of this exact reason and that's the only solution I can think of as well.

Oh, interesting.

> I would set the border-color for `.stationSearchButton:hover` and `:focus` to the same `#bdbdbd` as it is without interaction, if you're okay with that.

Sure!

MrKrisKrisu (approved):

PR #1416, opened 2023-02-21
:sparkles: add link to admin interface to status card

jeyemwey:
In the process, I've refactored how "is this user an admin" is checked in the blade files, pretty much hangling on [this stackoverflow answer](https://stackoverflow.com/a/54296004).

### DIFF
--- a/resources/views/includes/status.blade.php
+++ b/resources/views/includes/status.blade.php
@@ -194,6 +194,13 @@
                         </a>
                     </li>
                 @endif
+                @admin
+                    <li class="list-inline-item like-text">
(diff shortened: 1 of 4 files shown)


resources/views/includes/status.blade.php
```diff
@@ -194,6 +194,13 @@ class="like {{ $status->likes->where('user_id', auth()->user()->id)->first() ===
                         </a>
                     </li>
                 @endif
+                @admin
```
MrKrisKrisu, line 197: cool!

MrKrisKrisu (approved):

PR #1426, opened 2023-03-01
✨ display newlines in status body

### LINKED ISSUE #1349: Line Breaks for Checkin Descriptions instead of own Seperators
It would be nice if the Checkin Description allows Line Breaks. I write a lot of Informations in it and to seperate it I need to use |, I also saw others using the - to seperate it.

The Description is already a Textarea Input so it should be easy to allow the line breaks which can be made in Textareas so users don't need to use special seperators like - or | to seperate a sentence.

jeyemwey:
This fixes #1349. Ive moved to a PHP solution rather than a CSS solution since anything `pre` can cause weird ui bugs.

Since we have to use `{!!` in this case, I've added some more tests that HTML is properly sanitized.

In addition, a "only ever allow two linebreaks at a time" rule has been implemented in the UI, following [this stack exchange answer](https://stackoverflow.com/a/40084137):

image

### DIFF
--- a/resources/views/includes/status.blade.php
+++ b/resources/views/includes/status.blade.php
@@ -96,7 +96,7 @@
                     </p>
 
                     @if(!empty($status->body))
-                        <p class="status-body"><i class="fas fa-quote-right" aria-hidden="true"></i> {{ $status->body }}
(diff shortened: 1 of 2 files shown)


MrKrisKrisu (approved):
Good idea to limit the number of new lines.
I am a bit afraid with the `{!!`, but you've tested it. So... Okay for me.

PR #1427, opened 2023-03-01
:arrow_up: Upgrade to laravel 10

MrKrisKrisu:


MrKrisKrisu:
There is a package conflict with `uwumarie/laravel-bitflags` (https://github.com/NyCodeGHG/laravel-bitflags) pending. Maybe we wanna remove this dependency to reduce amount of packages.

/cc @NyCodeGHG 

```
  Problem 1
    - Root composer.json requires uwumarie/laravel-bitflags dev-main -> satisfiable by uwumarie/laravel-bitflags[dev-main].
    - uwumarie/laravel-bitflags dev-main requires illuminate/support ^9 -> satisfiable by illuminate/support[v9.0.0, ..., v9.52.7].

```

NyCodeGHG:
That repo only exists because the original package hasn't been updated, we could just merge the contents into here i think and get rid of the dependency?

MrKrisKrisu:
> That repo only exists because the original package hasn't been updated, we could just merge the contents into here i think and get rid of the dependency?

Hmm, I see. I'm not entirely happy with copying code from other repos into ours, but I will do this for now.
I don't see a license in this repo - I will add an license note that this isn't our code.
Maybe there is a better way for bitflags without external dependencies in the future.

//cc @jeyemwey @HerrLevin

PR #1495, opened 2023-03-24
:bug: Revert "Add button to 'join' other check ins"

### LINKED ISSUE #1491: Max Char Count on Checkin Status not showing.
**Describe the bug**
Max Char Count on Checkin Status not showing.

**To Reproduce**

1. Go to 'Trävelling'
2. Check in a new Train or edit a Check In
3. See error

**Expected behavior**
The Counter should count up or down when adding characters.

**Screenshots**
grafik

All Desktops and Phones with Firefox

MrKrisKrisu:
fixes #1491


ChillLP:
Why tho

MrKrisKrisu:
> Why tho

This Feature is causing some bugs. It is not possible to edit the statuses correctly because the modals are shown multiple times. Also there are colliding dom IDs when an edit-modal and create-modal are on the same page. This needs to be fixed before merging this again.

HerrLevin:
@MrKrisKrisu Can we fix this instead of removing it completely?

MrKrisKrisu:
> @MrKrisKrisu Can we fix this instead of removing it completely?

I can't fix this. Feel free to do that.

xanderio (approved):

PR #1520, opened 2023-04-01
fix: status interactions layout and like functionality

vainamov:
I fixed the layout for the interaction list, the items are now correctly aligned in a row.

I removed the `like` class from the login link, that replaces the like button when the current user isn't authenticated. This way the like function isn't called on clicked that link and the authentication checks in said function are no longer necessary. While at it, the current like state is now determined by a data-attribute instead of the Font Awesome classes of the icon, to make sure this won't backfire, if the icon style or library ever changes.


MrKrisKrisu:
Thanks for fixing the layout! But now you can no longer favorite a status. 🫣

PR #1569, opened 2023-05-05
:children_crossing: Improve the workflow to delete own profile picture

MrKrisKrisu:


app/Http/Controllers/API/v1/SettingsController.php
```diff
@@ -162,14 +161,26 @@ public function updateSettings(Request $request): UserProfileSettingsResource|Js
...
 
+    /**
+     * Undocumented and unofficial API Endpoint
+     *
+     * @return JsonResponse
+     */
     public function deleteProfilePicture(): JsonResponse {
         if (BackendSettingsController::deleteProfilePicture(user: auth()->user())) {
-            return $this->sendResponse('', 204);
+            return $this->sendResponse(['message' => __('settings.profilePicture.deleted')]);
```
HerrLevin, line 171: Http code 204 missing

HerrLevin (changes requested):

app/Http/Controllers/API/v1/SettingsController.php
MrKrisKrisu: changed to 200 because there is content now

MrKrisKrisu: ( 200 = default = missing declaration )

MrKrisKrisu: image

PR #1574, opened 2023-05-07
:sparkles: Digital Wellbeing: Disable Likes

### LINKED ISSUE #1515: Disable visibility of likes

In a discord conversation it came along that there are competitions between users regarding the count of likes (who has more likes). To ensure mental health, it should be possible to disable the visibility of likes.

There should be a option in the profile settings to disable the likes entirely, so you are not seeing any likes and of course not being able to like journeys. This also includes the notifications, so you don't get notified.  

**Additional context**

image

jeyemwey:
With this change, we're adding the option to disable the like function, which resolves #1515.

**When you have likes disabled**
* You will not see the Likes UI (the likes buttons and the tumblresk-history) at traewelling.de in any status.
* You will not see _new_ notifications for likes
* The API will always return that nobody has liked any status you encounter

**When you encounter someone with likes disabled**
* You will not see the Likes UI for that user's status both on traewelling.de and the API.
* You can not like those statuses via the API (returns 403).

** Changes in the UI**

There's a new setting in the settings plane:
image


app/Models/Status.php
```diff
@@ -63,6 +63,22 @@ public function getFavoritedAttribute(): ?bool {
...
+        if (!$this->user->likes_enabled) {
+            return false;
+        }
+
+        if (!Auth::check()) {
+            return true;
+        }
+
+        return Auth::user()->likes_enabled;
+    }
```
MrKrisKrisu: Why do you create two redundant methods? Isn't the model property enough?

MrKrisKrisu: (Maybe it would be better to move this logic to an Policy.)

database/migrations/2023_05_07_135454_add_likes_enabled_field_to_users_table.php
```diff
@@ -0,0 +1,35 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
```
MrKrisKrisu:
```suggestion
    public function up(): void
```

database/migrations/2023_05_07_135454_add_likes_enabled_field_to_users_table.php
```diff
@@ -0,0 +1,35 @@
...
+                  ->index();
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
```
MrKrisKrisu:
```suggestion
    public function down(): void
```

MrKrisKrisu (changes requested):
(all non blocking - thanks!)

app/Models/Status.php
jeyemwey: Kann man schon machen, ich hab nur keine Ahnung von Policies. Aber tu dir keinen Zwang an c:

> Isn't the model property enough?

Das property ist manchmal nicht evaluiert, wenn ich das im Code abgefragt habe, sondern war dann auf `null`, darum lieber so. Aber kann auch Jannik-Fettfingering sein.

MrKrisKrisu: Okay, ich probiere mich mal!

MrKrisKrisu (approved):
@jeyemwey Please have a look at my changes. :) If they are okay for you feel free to merge!

database/migrations/2023_05_07_135454_add_likes_enabled_field_to_users_table.php
```diff
@@ -0,0 +1,23 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void {
+        Schema::table('users', static function(Blueprint $table) {
+            $table->boolean('likes_enabled')
+                  ->default(true)
+                  ->after('shadow_banned')
+                  ->index();
```
MrKrisKrisu: oh. ähm. gerade noch gesehen. Wieso der Index? Danach searchen wir ja nicht. Würde ich ggfs. noch entfernen?

MrKrisKrisu (changes requested):

database/migrations/2023_05_07_135454_add_likes_enabled_field_to_users_table.php
MrKrisKrisu: (Jannik privat: "Ne der muss da nicht hin") -> entfernt

MrKrisKrisu (approved):

PR #1597, opened 2023-05-13
:coffin: Remove dead twitter integration

MrKrisKrisu:


MrKrisKrisu:
@jeyemwey @HerrLevin 

I would also like to clean up the database and delete the auth keys. I would still keep the account IDs so we can identify Twitter-only accounts at password reset. 

Can I delete this?

MrKrisKrisu:
:( * ping *

@jeyemwey @HerrLevin @xanderio

xanderio (approved):
Can't say much about the code, but I like the direction you're taking this. Do you have a plan when to remove the twitter user ids that are left in our database?

MrKrisKrisu:
> Can't say much about the code, but I like the direction you're taking this. Do you have a plan when to remove the twitter user ids that are left in our database?

I don't have a plan to delete them.
At the moment, users have the option to remove the association themselves in their account settings.
However, based on the current situation, I believe it's necessary to retain the user IDs for now. We're still receiving a significant number of requests from users who are unable to access their accounts due to not having an associated email address. Keeping the IDs allows us to assist them in recovering their accounts and maintaining a positive user experience.

HerrLevin:
Hm. I don't know if I want to remove it. I kinda wanted to re-implement twitter once they're clear about their API model and once I/we have figured out, if we can use their free API model.

xanderio:
We can always revert the change if and when twitter figures out if they want an actually useful free API tire. We should keep this code around just in case "elon suddenly gets sensible".

MrKrisKrisu:
Discuss with @HerrLevin:

Twitter code will be removed. Keys are still kept. We reserve the right to add a "Login with Twitter" again, but then with clean, new code.

Issue #647, opened 2021-12-14
[APIv1] Requesting notifications takes very long 


jheubuch:
**Describe the bug**
In APIv1 the endpoint /notifications [GET] is taking very long for loading the notifications (~30 secs)

**To Reproduce**

1. Request notifications via Postman or any other desired API client
2. Brew a cup of coffee ☕
3. See the result and the total request time

**Expected behavior**
Notifications should not take that long to load

**Screenshots**
![image](https://user-images.githubusercontent.com/60344409/145958457-f0420df1-8bd7-4e9b-a578-30f33988dcb8.png)

jheubuch:
Guess this is fixed by #1633

Thanks!

PR #1620, opened 2023-05-28
removed current notifications query limit

### LINKED ISSUE #1466: Notification Pill shouldn't stop at 10
If there are more than 10 Notifications, the Notification Pill will stop counting.
On the Beta, it displays more than 10 in the Pill.

It should not stop counting at 10 Notifications or it should display 9+ in the Pill.

tobiasreinhard:
Removed query limit from latest notifications, fixed #1466

### DIFF
--- a/app/Http/Controllers/NotificationController.php
+++ b/app/Http/Controllers/NotificationController.php
@@ -10,7 +10,6 @@ class NotificationController extends Controller
 {
     public static function renderLatest(): JsonResponse {
         $notifications = Auth::user()->notifications()
-                             ->limit(10)
                              ->get()
                              ->map(function($notification) {
                                  $notification->html = $notification->type::render($notification);


jeyemwey:
May I ask for your motivation for this? Some users have hundreds and thousands of notifications which would severely slow down their performance (fetching and rendering). That's why I'd first look into other options before just removing the limit.

MrKrisKrisu:
I understand the motivation behind it, I also sometimes have > 10 notifications and only see max. 10 pieces.

However, the programming of the notifications is not designed to load so many notifications, that would delay the loading extremely. (See the APIv1 implementation - the endpoint of it takes forever to load - Issue #647).

I would therefore close the PR until the performance of the query has been optimized. Sorry.

tobiasreinhard:
Sorry, i messed up. I thought it only loads current notifications of a user, not all notifications at once.

MrKrisKrisu:
I think the best way would be to add Pagination aroudn this and optimize the mapping with Eager Loading.

Thanks anyway for the PR! <3

PR #1648, opened 2023-06-08
:globe_with_meridians: remove stray "reason" from distance calculation text

evgeni:


MrKrisKrisu:
Thanks! :)

However, for easier translation changes, I would recommend using Weblate in the future. 
-> https://translate.codeberg.org/projects/trawelling/traewelling/

evgeni:
Heh. It was a random thing I noticed while surfing the site, and git was right there, ready to be used :)

PR #1665, opened 2023-06-12
🧑‍💻 Re-Implement Notifications w/ vue.js

HerrLevin:
This is just a simple re-implementation of the Notifications based on @MrKrisKrisu's original implementation but in vue3.

This reduces the complexity and need for DOM-Accessors.


MrKrisKrisu:
I can't give you a meaningful review because I don't know Vue.

resources/js/app.js
```diff
@@ -2,13 +2,30 @@
...
+const app = createApp({
+	data() {
+		return{
+			count: 0
+		}
+	}
+});
+
+app.component('notificationlist', NotificationList);
+app.component('notificationbell', NotificationBell);
```
vainamov: I might be wrong, but you're using these components in DOM templates so shouldn't these be registered as `NotificationList` and `NotificationBell` and then be referenced in _app.blade.php_ as `<notification-list>` and `<notification-bell>` respectively?

https://vuejs.org/guide/components/registration.html#component-name-casing

resources/vue/components/NotificationBell.vue
```diff
@@ -0,0 +1,110 @@
...
+        type: String
+    },
+    i18nTitle: {
+        type: String
+    }
+});
+
+onMounted(() => {
+    fetchCount();
+    setInterval(fetchCount, 30000);
```
vainamov: Intervals should be properly destroyed, see [onUnmounted()](https://vuejs.org/api/composition-api-lifecycle.html#onunmounted)

resources/vue/components/NotificationEntry.vue
```diff
@@ -0,0 +1,126 @@
...
+        },
+        read() {
+            return this.readAt ?? false;
+        }
+    },
+}
+</script>
+
+<template>
+    <div class="row notification" :class="[warnType, read ? '' : 'unread']">
```
vainamov:
```suggestion
    <div class="row notification" :class="[warnType, { unread: !read }]">
```

resources/vue/components/NotificationBell.vue
```diff
@@ -0,0 +1,110 @@
...
+        :hide-footer="true">
+        <template #body class="p-0">
+            <NotificationList
+                ref="notifications"
+                @toggle-read="fetchCount"
+                :empty-text="i18nEmpty"
+            />
+        </template>
+        <template #header-extra>
+            <a
```
vainamov: Semantically this should be a `<button>`

resources/vue/components/NotificationBell.vue
```diff
@@ -0,0 +1,110 @@
...
+        .then(function (request) {
+            request.json().then(function (json) {
+                state.count = json.data;
+            });
+        });
+}
+</script>
+
+<template>
+    <a @click="showModal" class="notifications-board-toggle"
```
vainamov: Semantically this should be a `<button>`

vainamov (commented):
I took the liberty of reviewing this, I hope it's okay 😅

resources/js/app.js
HerrLevin: afaik Vue resolves to every possible case-constellation, but kebab-case seems to be the "correct" way of doing it. Will do!

resources/vue/components/NotificationBell.vue
```diff
@@ -0,0 +1,119 @@
...
+        <template #body class="p-0">
+            <NotificationList
+                ref="notifications"
+                @toggle-read="fetchCount"
+                :empty-text="i18nEmpty"
+            />
+        </template>
+        <template #header-extra>
+            <button
+                href="#"
```
vainamov:
```suggestion
                type="button"
```

HerrLevin: 🤦🏻‍♂️ right, thanks

PR #1668, opened 2023-06-16
✨ Add new date constraints to events, to separate checkin dates from event dates

### LINKED ISSUE #1593: Make Events "longer"
For the Japantag in Düsseldorf For example, some of my Friends start träwelling today, so having the abitity to use the Event Tag a day before and after the Event with the Addition of "Anreise" (day before) "Abreise" day after would make Sense since some Events are very difficult to get to on one day

HerrLevin:
Closes #1593

### DIFF
--- a/app/Http/Controllers/Frontend/Admin/EventController.php
+++ b/app/Http/Controllers/Frontend/Admin/EventController.php
@@ -20,6 +20,19 @@ use Illuminate\View\View;
 
 class EventController extends Controller
 {
+
+    private const VALIDATOR_RULES = [
+        'name'                 => ['required', 'max:255'],
+        'hashtag'              => ['required', 'max:30'],
+        'host'                 => ['required', 'max:255'],
+        'url'                  => ['nullable', 'url'],
+        'nearest_station_name' => ['required', 'max:255'],
+        'begin'                => ['required', 'date'],
+        'end'                  => ['required', 'date'],
+        'event_start'          => ['required', 'date', 'after_or_equal:begin'],
+        'event_end'            => ['required', 'date', 'before_or_equal:end'],
+    ];
+
     public function renderList(Request $request): View {
         $events = Event::orderByDesc('end');
         if ($request->has('query')) {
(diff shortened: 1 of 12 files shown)


app/Models/Event.php
```diff
@@ -45,7 +49,11 @@ public function getTrainDurationAttribute(): int {
...
 
     public function getIsPrideAttribute(): bool {
-        $event_name_lowercase = strtolower($this->name);
-        return Str::contains($event_name_lowercase, ['csd', 'pride']);    
+        $eventNameLowercase = strtolower($this->name);
+        return Str::contains($eventNameLowercase, ['csd', 'pride']);
+    }
+
+    public function approvedBy(): HasOne {
+        return $this->hasOne(User::class, 'id', 'approved_by');
```
MrKrisKrisu, line 57: unused and always-null function?

app/Http/Controllers/Frontend/Admin/EventController.php
```diff
@@ -20,6 +20,19 @@
...
+    private const VALIDATOR_RULES = [
+        'name'                 => ['required', 'max:255'],
+        'hashtag'              => ['required', 'max:30'],
+        'host'                 => ['required', 'max:255'],
+        'url'                  => ['nullable', 'url'],
+        'nearest_station_name' => ['required', 'max:255'],
+        'begin'                => ['required', 'date'],
+        'end'                  => ['required', 'date'],
+        'event_start'          => ['required', 'date', 'after_or_equal:begin'],
+        'event_end'            => ['required', 'date', 'before_or_equal:end'],
```
MrKrisKrisu: Hmm, i don't see a reason why we need to make all of this required.

Imo we just need `event_start` and `event_end` required, begin and end is optionally. Most of events doesn't need another date for showing them.

MrKrisKrisu (changes requested):

database/migrations/2023_06_13_213444_expand_event_date_restrictions_and_add_fulltext_search.php
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+
+    public function down(): void
+    {
+        Schema::table('events', static function (Blueprint $table) {
+            $table->dropColumn('event_start');
+            $table->dropColumn('event_end');
+            $table->dropForeign('events_approved_by_foreign');
+            $table->dropColumn('approved_by');
```
MrKrisKrisu:
```suggestion
            $table->dropColumn(['event_start', 'event_end', 'approved_by']);
```

should work. (please test)

resources/views/admin/events/suggestions.blade.php
```diff
@@ -1,3 +1,4 @@
+@php use Illuminate\Support\Facades\Date; @endphp
```
MrKrisKrisu, line 1: do you like this? I try to don't import things in blade, because I find it ugly. 👀

MrKrisKrisu (changes requested):

app/Models/Event.php
HerrLevin: Yes, but I kinda need it in an upcoming feature. Either I remove it now and re-add it tomorrow, or we leave it here. 😅

resources/views/admin/events/suggestions.blade.php
HerrLevin: Yes, I like this. I hate FQDNs. :D

MrKrisKrisu (approved):

PR #1670, opened 2023-06-16
✨ Add reasons for event suggestion rejection

HerrLevin:


app/Enum/EventRejectionReason.php
```diff
@@ -0,0 +1,15 @@
...
+namespace App\Enum;
+
+enum EventRejectionReason: string
+{
+    case DEFAULT = 'denied';
+    case LATE = 'too-late';
+    case DUPLICATE = 'duplicate';
+    case NOT_APPLICABLE = 'not-applicable';
+
+    public function getReason(): string {
```
MrKrisKrisu, line 12: that doesn't return the reason? Maybe rename the function to `getTranslationKey` or similar?

app/Http/Controllers/Frontend/Admin/EventController.php
```diff
@@ -54,17 +56,22 @@ public function renderEdit(int $id): View {
     }
 
     public function denySuggestion(Request $request): RedirectResponse {
-        $validated       = $request->validate(['id' => ['required', 'exists:event_suggestions,id']]);
+        $validated = $request->validate([
+                                            'id'      => ['required', 'exists:event_suggestions,id'],
+                                            'decline' => ['required', new Enum(EventRejectionReason::class)]
```
MrKrisKrisu: `decline` is hard to understand. maybe something like `declineReason`?

app/Notifications/EventSuggestionProcessed.php
```diff
@@ -11,12 +12,14 @@ class EventSuggestionProcessed extends Notification implements BaseNotification
 {
     use Queueable;
 
-    private EventSuggestion $eventSuggestion;
-    private ?Event          $event;
+    private EventSuggestion      $eventSuggestion;
+    private ?Event               $event;
+    private ?EventRejectionReason $reason;
```
MrKrisKrisu: please format correctly

app/Notifications/EventSuggestionProcessed.php
```diff
@@ -11,12 +12,14 @@ class EventSuggestionProcessed extends Notification implements BaseNotification
...
 
-    public function __construct(EventSuggestion $eventSuggestion, ?Event $event) {
+    public function __construct(EventSuggestion $eventSuggestion, ?Event $event, ?EventRejectionReason $reason = null) {
```
MrKrisKrisu:
```suggestion
    public function __construct(EventSuggestion $eventSuggestion, ?Event $event, EventRejectionReason $reason = null) {
```

don't need to be nullable if the default is null.

app/Notifications/EventSuggestionProcessed.php
```diff
@@ -28,6 +31,7 @@ public function toArray(): array {
             'accepted'      => $this->event !== null,
             'event'         => $this->event?->only(['id', 'slug', 'name', 'begin', 'end']),
             'suggestedName' => $this->eventSuggestion->name,
+            'reason'        => $this->reason,
```
MrKrisKrisu: `declineReason` to be consistent.

resources/views/admin/events/suggestion-create.blade.php
```diff
@@ -73,6 +74,42 @@ class="form-control" placeholder="Closest Träwelling station"
...
+                                </button>
+                            </li>
+                            <li>
+                                <button class="btn-link dropdown-item" name="decline"
+                                        value="{{EventRejectionReason::DEFAULT}}">No Reason
+                                </button>
+                            </li>
+                        </ul>
+                    </div>
+                </div>
```
MrKrisKrisu: please outsource this code duplication with `resources/views/admin/events/suggestions.blade.php` (57-84) to one place.

MrKrisKrisu (changes requested):

app/Enum/EventRejectionReason.php
MrKrisKrisu: or remove this function. I don't see any reason to have this, as of the string is only used in one function... therefore we can reduce complexivity

app/Enum/EventRejectionReason.php
```diff
@@ -10,6 +10,6 @@ enum EventRejectionReason: string
     case NOT_APPLICABLE = 'not-applicable';
 
     public function getReason(): string {
-        return sprintf('notifications.eventSuggestionProcessed.%s', $this->value);
+        return __(sprintf('notifications.eventSuggestionProcessed.%s', $this->value));
     }
```
MrKrisKrisu: Maybe remove this function? I don't see any reason to have this, as of the string is only used in one function... therefore we can reduce complexivity.

app/Http/Controllers/Frontend/Admin/EventController.php
```diff
@@ -57,8 +57,8 @@ public function renderEdit(int $id): View {
 
     public function denySuggestion(Request $request): RedirectResponse {
         $validated = $request->validate([
-                                            'id'      => ['required', 'exists:event_suggestions,id'],
-                                            'decline' => ['required', new Enum(EventRejectionReason::class)]
+                                            'id'            => ['required', 'exists:event_suggestions,id'],
+                                            'declineReason' => ['required', new Enum(EventRejectionReason::class)]
```
MrKrisKrisu: Sorry, I'm just seeing it now. Do we call it `decline` or `reject`? Because the Enum is called `EventRejectionReason`.

app/View/Components/EventDeclineButton.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace App\View\Components;
+
+use Illuminate\View\Component;
+
+class EventDeclineButton extends Component
+{
+    public function render()
```
MrKrisKrisu: return type?

MrKrisKrisu (changes requested):

app/View/Components/EventDeclineButton.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace App\View\Components;
+
+use Illuminate\View\Component;
+
+class EventDeclineButton extends Component
```
MrKrisKrisu: Also here: Decline or Reject?

MrKrisKrisu (changes requested):

app/Enum/EventRejectionReason.php
HerrLevin: imho it's cleaner to leave it herre. This way, we could change the values of the return types and still return the correct translation instead of guessing in several positions.

app/Http/Controllers/Frontend/Admin/EventController.php
HerrLevin: oh boy. I wanna go the easy route and rename the Enum, but reject is the better fitting word... 🥲

PR #1697, opened 2023-06-26
:sparkles: Add categories to events

### LINKED ISSUE #1689: Add categories to events
To de-clutter the complexity of events & the event overview page, Events should be able to have categories. That way, one can filter for pride events, festivals, Special train rides, etc.

**Additional context**
We should statically define the events with an enum and the corresponding translations rather than database-managed. That way, we will always have translations w/ a PR rather than possible new categories, that are not translatable by us/weblate.

MrKrisKrisu:
closes #1689 

This PR adds an backend and admin frontend for event categorization.


MrKrisKrisu:
no motivation to continue working on this - if you want you can continue this @HerrLevin

PR #1698, opened 2023-06-26
🐛 Fix timezone checkins

### LINKED ISSUE #1693: Cannot check-in if there are no stations, also not to last/end station
**Describe the bug**
When checking in from Wigan North Western to  Liverpool Lime Street (a Great Britian "Northern line" train), you cannot check-in, because:
* no stations are listed
* but you also cannot check-in to the last station via that dedicated button unfortunately, which I don't get why that does not work 

**To Reproduce**

1. Go to https://traewelling.de/trains/stationboard?provider=train&ibnr=7002411&when=2023-06-25T17%3A18%3A11%2B02%3A00 and check-in in 17:23 – in UTC+02, but [that is another topic](https://github.com/Traewelling/traewelling/issues/1692)):
   grafik
2. So there are no stations [on the next page](https://traewelling.de/trains/trip?tripID=1%7C155208%7C4%7C80%7C25062023&lineName=59450&start=7002411&departure=2023-06-25T17%3A23%3A00%2B01%3A00):
   grafik
3. Click on that special button at the top right that I expect may be made for this purpose or in general when for going to the last station:
   grafik
4. You get an error:
   grafik

**Expected behavior**
Should check-in to last station, I mean it has the correct name, should not that be possible?

**Screenshots**
network traffic causing error:
grafik

**Desktop (please complete the following information):**
- OS: Linux/Fedora
- Browser: Firefox 
- Version: 114

**Smartphone (please complete the following information):**
(also reproduced)

**Additional context**
N/A

HerrLevin:
Timezone-checkins are still very much broken but at least one can now check in. Closes #1693

### DIFF
--- a/app/Http/Controllers/FrontendTransportController.php
+++ b/app/Http/Controllers/FrontendTransportController.php
@@ -119,7 +119,14 @@ class FrontendTransportController extends Controller
 
         try {
             $startStation = TrainStation::where('ibnr', $validated['start'])->firstOrFail();
-            $departure    = Carbon::parse($validated['departure']);
+
+            // If we get a string with a timezone, remove it, so we can parse it "correctly"
+            // This needs to be removed once we can handle UTC and timezones.
+            if (str_contains(substr($validated['departure'], -6), '+')) {
+                $departure = Carbon::parse(substr($validated['departure'], 0, -6));
+            } else {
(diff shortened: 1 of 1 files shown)


MrKrisKrisu (approved):
hmmmmm... okay. but.. we need to improve that.

giphy

PR #1701, opened 2023-07-02
:zap: Use async decoding and lazy loading for profile pictures

AlexMeier99:
Hi,  
I just discovered this great project. I'm a web developer myself and a big fan of public transportation in general.  
Apparently this project currently does not use [lazy loading](https://caniuse.com/loading-lazy-attr) or [async decoding](https://caniuse.com/mdn-html_elements_img_decoding) for the profile pictures.  
Applying both should reduce the server load on pages with many `image-box` elements such as <https://traewelling.de/statuses/active> or the user profile pages.

Currently all images are loaded as `png` files in **parallel** from the server. By adding the `loading=lazy` html attribute the browser will load the images when needed (before they appear in the viewport). The async decoding should further speed up the rendering (please also see [here](https://css-tricks.com/newsletter/249-decoding-async-tree-rings-and-flexbox-gap/)).

Using modern media formats like [`webp`](https://caniuse.com/webp) or [`avif`](https://caniuse.com/avif) could further decrease the file sizes. Using the [default](https://traewelling.de/img/user.png) profile picture as an example the file size could be decreased from 16.4kb to 4.1kb when encoding the image as `webp`.

It's also worth mentioning that [optimizing](https://css-tricks.com/tools-for-optimizing-svg/) the `svg` icons (tram, suburban, etc.) with a tool like [SVGOMG](https://jakearchibald.github.io/svgomg/) or [svgo](https://github.com/svg/svgo) might also be feasable.

Best regards,
Alex

### DIFF
--- a/resources/views/includes/status.blade.php
+++ b/resources/views/includes/status.blade.php
@@ -24,7 +24,7 @@
     <div class="card-body row">
         <div class="col-2 image-box pe-0 d-none d-lg-flex">
             <a href="{{ route('profile', ['username' => $status->user->username]) }}">
-                <img src="{{ ProfilePictureController::getUrl($status->user) }}"
+                <img loading="lazy" decoding="async" src="{{ ProfilePictureController::getUrl($status->user) }}"
                      alt="{{ $status->user->username }}">
             </a>
         </div>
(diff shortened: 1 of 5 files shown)


MrKrisKrisu (approved):
Thanks for this changes! Maybe u can create issues for your other suggestions, if you don't wanna make a PR for this, too?
Then it is on our list and anyone can be do this some day™.

Issue #1718, opened 2023-07-07
Cannot check in some UK trains


timoschwarzer:
**Describe the bug**
I tried to check in to this train: https://traewelling.de/trains/trip?tripID=1%7C156441%7C0%7C80%7C7072023&lineName=59347&start=7000431&departure=2023-07-07T21%3A00%3A00%2B01%3A00 resulting in an error message ("An unknown error occurred. Please try again.")

**To Reproduce**

1. Click link above
2. Select "Peterborough"

**Desktop (please complete the following information):**

- OS: Linux
- Browser Chrome
- Version 114

HerrLevin:
This should be fixed with https://github.com/Traewelling/traewelling/pull/1719 and rolled out on prod

PR #1719, opened 2023-07-07
🐛 Fix Timezones in a very big way

### LINKED ISSUE #1692: Show trains in local time of station or browser instead of UTC+01/UTC+02 (MEZ)
So I wanted to checkin to a train in Glasgow as I am there. This is UTC+01 (British summer time), further called the local time, apparently.

So according to local time the train starts 13:35.
However, Trawelling uses the German/Berlin time (UTC+02 currently as German summer time), so in Traewelling the train starts at 14:35.

grafik
https://traewelling.de/status/1153504

Either:
* always evaluate the local timezone of the browser and adjust all times to it (I've shown that the time zone is correct in my screenshot)
* or do that depending on the start station you select (as you are very obviously there it is the time that matters for you)
* or do some of that, but only show a warning – e.g. show a warning when the browser time zone does not match Traewellings time zone – that you may be traewelling in a different time zone and may adjust/calculate the correct time on your head

Dunno, listed above. IMHO what matters when searching trains is the _local time_ I am taking the train as that is what I have in my head and is shown everywhere.
Leaving it as it is may confuse people. I don#t think most people think as technical as remembering that Traewelling (and the Deutsche Bahn API it is fed of) are located in Germany and you would have to use these times. And even if so, there are many ideas to make it obvious a time zone adjustment has been done, e.g.:
* show the time zone explicitly near the time, e.g. "13:35 (UTC+01)",  – a tooltip may show "British summertime" even or 
* show a warning or information message somewhere that time has been adjusted

**Additional context**
Found https://github.com/Traewelling/traewelling/issues/6 being related.

If you need a good JS DateTime library I can recommend [Luxon](https://moment.github.io/luxon/#/) as it is a wrapper around the JS `DateTime` object.

HerrLevin:
This should hopefully once and for all fix the general issues with Timezoning. (Users hosting their own instance of Träwelling should change the timezone of their database to UTC, otherwise this will create unexpected behavior)

Fixes #1692 #1718

image

image

Europe/Berlin:
image

Europe/London:
image

Pacific/Kwajalein:
image


app/Casts/UTCDateTime.php
```diff
@@ -0,0 +1,45 @@
...
+     * @throws Exception
+     */
+    public function set($model, string $key, $value, array $attributes): Carbon|null
+    {
+        if ($value === null) {
+            return null;
+        }
+        if (!$value instanceof Carbon) {
+            if (is_string($value) && !str_contains($value, '+')) {
+                throw new Exception("THIS IS A WONKY LOCAL TIME!!1 -> $value");
```
MrKrisKrisu:
```suggestion
                throw new Exception("Given timestamp has no valid timezone in it -> $value");
```

app/Helpers/Helper.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use Illuminate\Support\Carbon;
```
MrKrisKrisu, line 3: Please use eather Carbon\Carbon or Facades\Date.

app/Helpers/Helper.php
```diff
@@ -40,3 +42,12 @@ function durationToSpan($duration): string {
 
     return $return;
 }
+
+function userTime(null|Carbon|\Carbon\Carbon|string $time=null, ?string $format=null, bool $iso=true): string {
```
MrKrisKrisu, line 46: see review comment above in line 3

```suggestion
function userTime(null|\Carbon\Carbon|string $time=null, ?string $format=null, bool $iso=true): string {
```

app/Helpers/Helper.php
```diff
@@ -40,3 +42,12 @@ function durationToSpan($duration): string {
...
+    $format = $format ?? __('time-format');
+    $time = $time instanceof \Carbon\Carbon ? $time : Carbon::parse($time);
+    if ($iso) {
+        return $time->tz(auth()->user()->timezone)->isoFormat($format);
+    }
+    return $time->tz(auth()->user()->timezone)->format($format);
```
MrKrisKrisu, line 52: is the timezone attribute nullable? If so please catch

MrKrisKrisu: why not isoFormat everytime?

app/Http/Controllers/API/v1/StatusController.php
```diff
@@ -22,6 +22,7 @@
 use Illuminate\Http\Request;
 use Illuminate\Http\Resources\Json\AnonymousResourceCollection;
 use Illuminate\Http\Response;
+use Illuminate\Support\Carbon;
```
MrKrisKrisu, line 25: Please use eather Carbon\Carbon or Facades\Date.

app/Http/Controllers/Backend/Transport/TrainCheckinController.php
```diff
@@ -136,10 +136,10 @@ private static function createTrainCheckin(
 
         //Note: Compare with ->format because of timezone differences!
         $firstStop = $trip->stopovers->where('train_station_id', $origin->id)
-                                     ->where('departure_planned', $departure->format('Y-m-d H:i:s'))
+                                     ->where('departure_planned', $departure)
```
MrKrisKrisu, line 139: Is this really working? I had issues with this in the past in this function.

MrKrisKrisu (changes requested):

app/Casts/UTCDateTime.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace App\Casts;
+
+use Exception;
+use Illuminate\Contracts\Database\Eloquent\CastsAttributes;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Carbon;
+
+class UTCDateTime implements CastsAttributes
```
MrKrisKrisu, line 18: Can you add a PHPDoc where you explain the usage of this Cast, so we know in future?

MrKrisKrisu (changes requested):

app/Helpers/Helper.php
HerrLevin: I'm not entirely sure about that. They explicitly state this in their documentation and the returntype of a `datetime`-cast is also `Illuminate\Support\Carbon`. https://laravel.com/docs/10.x/helpers#dates

app/Helpers/Helper.php
HerrLevin: Because isoFormat is only used in our translations (I'm not really sure why) and in my opinion, `->format()` is generally the cleaner version.

(Also we're using 90% regular date formatting. `Y-m-d H:i:s` is just better than `YYYY-MM-DD HH:mm:ss` :D )

app/Http/Controllers/Backend/Transport/TrainCheckinController.php
HerrLevin: It's not not working. Laravel will always cast datetime-castables to `Y-m-d H:i:s` inside its query builder. If you're supplying a datetimestring, the query builder will parse it to carbon and then render it to `Y-m-d H:i:s`.

On top of that, `departure_planned` has been changed to UTCDateTime, which needs an instance of carbon or a datetime string with a timezone, otherwise it will fail with "THIS IS A WONKY LOCAL TIME!!1" :D

You'll still need `->format('Y-m-d H:i:s')` in a `whereRAW`-condition but as long as it's passing through the query builder, we should supply the Carbon model. (This also saves RAM!!1 :D)

app/Helpers/Helper.php
```diff
@@ -46,8 +46,9 @@ function durationToSpan($duration): string {
 function userTime(null|Carbon|\Carbon\Carbon|string $time=null, ?string $format=null, bool $iso=true): string {
     $format = $format ?? __('time-format');
     $time = $time instanceof \Carbon\Carbon ? $time : Carbon::parse($time);
+    $timezone = auth()->user()->timezone ?? config('app.timezone');
```
MrKrisKrisu, line 49: timezone can't be null, because there is a db default

HerrLevin: can be null, because `auth()->user()` can be null.

PR #1751, opened 2023-07-18
:sparkles: Show users also on this connection

### LINKED ISSUE #1750: See who Else is checked in
After checking in, you only get a small Windows showing who is checked in, and after that is gone its rather difficult seeing who Else is there. 

I Imagine after pulling Up your own checking, that below that there will be a section of the other people that are on that Ride, a little Bit Like a "you might also Like" Section

I promise i'll be done soon with Feature ideas

MrKrisKrisu:
fixes #1750

I hate to design a UI - has anyone some suggestions how to improve this? I don't really like this try.

image

(and yes @jheubuch, I'll create an API Endpoint for this)

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -43,7 +43,7 @@ function durationToSpan($duration): string {
     return $return;
 }
 
-function userTime(null|Carbon|\Carbon\Carbon|string $time=null, ?string $format=null, bool $iso=true): string {
+function userTime(null|Carbon|\Carbon\Carbon|string $time = null, ?string $format = null, bool $iso = true): string {
     if ($time === null) {
         return '';
     }
--- a/app/Models/Status.php
+++ b/app/Models/Status.php
@@ -17,7 +17,7 @@ use Illuminate\Support\Facades\Auth;
  * @property Business         business
  * @property int              event_id
  * @property StatusVisibility visibility
- * @property TrainCheckin $trainCheckin
+ * @property TrainCheckin     $trainCheckin
  */
 class Status extends Model
 {
--- a/app/Models/TrainCheckin.php
+++ b/app/Models/TrainCheckin.php
@@ -43,7 +43,7 @@ class TrainCheckin extends Model
         'distance', 'duration', 'departure', 'real_departure', 'arrival', 'real_arrival', 'points', 'forced',
     ];
     protected $hidden   = ['created_at', 'updated_at'];
-    protected $appends  = ['origin_stopover', 'destination_stopover', 'speed', 'alsoOnThisConnection'];
+    protected $appends  = ['origin_stopover', 'destination_stopover', 'speed', 'alsoOnThisConnection', 'stopovers'];
     protected $casts    = [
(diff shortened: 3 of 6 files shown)


jheubuch:
Can we have this data in the "default" /v1/status/123456 endpoint? (Or at least if there are co-travellers to be requested)

Because I just want to request the co-travellers if I am sure that there are any (aka I wanna save requests)

Do you understand what I mean? :)

MrKrisKrisu:
> Can we have this data in the "default" /v1/status/123456 endpoint?

Sure, I don't wanna create an additional endpoint. But I wanna make this optional - maybe with an additional GET-Parameter like `v1/status/123456?showCoTravellers=true` or similar?

jheubuch:
> > Can we have this data in the "default" /v1/status/123456 endpoint?
> 
> Sure, I don't wanna create an additional endpoint. But I wanna make this optional - maybe with an additional GET-Parameter like `v1/status/123456?showCoTravellers=true` or similar?
> 

Yes! Nice!

HerrLevin:
I like the feature but I don't really like that we're stacking new items in the same column.
This will be particularly hard to read once a few people have liked the checkin.

Normally I wouldv'e made a small mockup but I'm on vacation so you'll have to deal with my small sketch here. :D

I think the best compromise would be, to introduce a second column on the right. (We could also use it to show the operator if we have one present)

To reduce complexity, we should only show the profile picture of the checked in users. Once hovering, we should show a popover containing the username and their start and end of the journey.

Clicking on their profile picture should redirect to the status.
image

image
(please ignore the ugly hand 🥴)

ChillLP:
Making the collumn of people who Liked into an Pop-up instead wouldnt be an Option?

HerrLevin:
> Making the collumn of people who Liked into an Pop-up instead wouldnt be an Option?

Also a possibility but either way I'd still prefer the "also on this connection" information on the right.

ChillLP:
To be fair, the page is generally quiete empty.
Maybe it could be devided into 3 or 4 sections...

lemme sketch something real quick

ChillLP:
So, two versions i crappy made in Photoshop Rn

Unbenannt-1

Unbenannt-2

ChillLP:
The space For Comments could also be used For some information of the current Ride If available, For example Max Speed, Baureihe, Wagenreihung, EVU and so on

MrKrisKrisu:
image

just a little sketch - I would like it, if we can maybe show the logo of the operator on the top instead of this dummy text.
Maybe we can make a wikidata connection or do something with crowdsourcing to get the logo's?

And yes, there are some spaces missing - just a little sketch for the features.

ChillLP:
Honestly i love the idea

MrKrisKrisu:
Maybe someone else would like to take this up again? My expertise with designing UI is depressing.

ChillLP:
sad, i found your idea quite nice

ChillLP:
Looking back, maybe that design could  - below/or if no others in Connection show the Tags the user set or that Journey

Issue #1686, opened 2023-06-20
Correct database field name for better dev-experience


HerrLevin:
**Describe the bug**
The newly added field of #1439 are a bit ambiguous.
They are named `real_departure` and `real_arrival` which could be misunderstood with actual HAFAS realtime Data. 

They should be renamed to something like `manual_departure`, so that their use is completely clear

Issue #1043, opened 2022-09-11
[API] Add API Documentation for V1


HerrLevin:
- [x]     Route::group(['prefix' => 'auth'], function() {
 - [x]         Route::post('login', [v1Auth::class, 'login']);
 - [x]         Route::post('signup', [v1Auth::class, 'register']);
 - [x]         Route::group(['middleware' => 'auth:api'], function() {
 - [x]             Route::post('refresh', [v1Auth::class, 'refresh']);
 - [x]             Route::post('logout', [v1Auth::class, 'logout']);
 - [x]             Route::get('user', [v1Auth::class, 'user']);
 - [x]     Route::get('static/privacy', [PrivacyPolicyController::class, 'getPrivacyPolicy'])

Route::group(['middleware' => ['auth:api', 'privacy-policy']], function() {
 - [x]         Route::post('event', [EventController::class, 'suggest']);
 - [x]         Route::get('activeEvents', [EventController::class, 'activeEvents']);
 - [x]         Route::get('leaderboard/friends', [StatisticsController::class, 'leaderboardFriends']);
 - [x]         Route::get('dashboard', [StatusController::class, 'getDashboard']);
 - [x]         Route::get('dashboard/global', [StatusController::class, 'getGlobalDashboard']);
 - [x]         Route::get('dashboard/future', [StatusController::class, 'getFutureCheckins']);
 - [x]         Route::post('like/{statusId}', [LikesController::class, 'create']);
 - [x]         Route::delete('like/{status}', [LikesController::class, 'destroy']);
 - [x]         Route::delete('statuses/{id}', [StatusController::class, 'destroy']);
 - [x]         Route::put('statuses/{id}', [StatusController::class, 'update']);
 - [ ]         Route::post('support/ticket', [SupportController::class, 'createTicket']);
 
Route::group(['prefix' => 'notifications'], function() {
 - [ ]             Route::get('/', [NotificationsController::class, 'index']);
 - [ ]             Route::get('count', [NotificationsController::class, 'count']);
 - [ ]             Route::put('{id}', [NotificationsController::class, 'update']);
 - [ ]             Route::put('read/{id}', [NotificationsController::class, 'read']);
 - [ ]             Route::put('unread/{id}', [NotificationsController::class, 'unread']);
 - [ ]             Route::post('readAll', [NotificationsController::class, 'readAll']);

Route::group(['prefix' => 'trains'], function() {
[...]

PR #1802, opened 2023-08-05
:memo: Change description of api docs

MrKrisKrisu:
Since the documented endpoints are relatively stable I suggest to change the description as in the PR.

Please have a look and approve/review this @HerrLevin

### DIFF
--- a/app/Http/Controllers/API/v1/Controller.php
+++ b/app/Http/Controllers/API/v1/Controller.php
@@ -6,11 +6,9 @@ use Illuminate\Http\JsonResponse;
 
 /**
  * @OA\Info(
- *      version="1.0.0 - alpha",
+ *      version="1.0.0",
  *      title="Träwelling API",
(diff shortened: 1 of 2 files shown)


HerrLevin:
I'm against this change, because there are still some endpoints missing and #1686 is still not done, which will be a breaking change in the API.

HerrLevin:
(See also #1043 - once all is done there, the API can move out of the alpha.)

MrKrisKrisu:
> I'm against this change, because there are still some endpoints missing and https://github.com/Traewelling/traewelling/issues/1686 is still not done, which will be a breaking change in the API.

Good points. I'll wait.

> See also https://github.com/Traewelling/traewelling/issues/1043 - once all is done there, the API can move out of the alpha.

Hmm. I don't accept that... The API can be marked as "ready" even if there are missing endpoints.

PR #1810, opened 2023-08-06
:children_crossing: Redirect to status after checkin

MrKrisKrisu:
This will improve the user experience as of this page loads a little bit faster.


ChristianTacke:
Thanks! This improves UX a lot. I often use the destination of one checkin to start the next checkin (in a multistep journey).

PR #1857, opened 2023-08-18
Add nix development environment

NyCodeGHG:
This PR adds a local development environment using [nix](https://nixos.org) and [devenv](https://devenv.sh).
I documented on how to use it in the readme.
This makes getting started with hacking on träwelling way easier and faster.
Of course it's not required and completely optional, but since i use this and it's always a bit annoying to watch out to not commit this and at least one other user on the träwelling discord uses devenv, i thought it would be good to have this in the repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,7 @@ public/js/*.js.LICENSE.txt
 public/css/*.css
 public/css/*.css.map
 .phpunit.result.cache
+
+# Nix
+.devenv
+.direnv
--- a/.gitignore
+++ b/.gitignore
@@ -137,3 +137,4 @@ public/css/*.css.map
 # Nix
 .devenv
 .direnv
+result*
--- a/README.md
+++ b/README.md
@@ -86,6 +86,23 @@ php artisan passport:install
 Use your webserver of choice or the in php included dev server (`php artisan serve`) to boot the application.
 You should see the Träwelling homepage at http://localhost:8000.
 
+### Option 3: Local Development using [Nix](https://nixos.org/)
+
+Nix is a cross-platform package manager for Linux and macOS systems.
+It also provides per project development environments.
+There is a also a Linux Distribution called NixOS which builds on top of nix,
+but it's not required to use nix and this development environment.
+
+Our [nix flake](flake.nix) includes such an environment with a pre configured MySQL instance.
+
+If you want to use it:
+
+- [Install nix](https://github.com/DeterminateSystems/nix-installer) if you haven't already. (Make sure you've [enabled flakes and the nix command](https://nixos.wiki/wiki/Flakes#Permanent))
+- Activate the environment either by using the [direnv](https://direnv.net/) shell hook or by executing `nix develop --impure` ([why-impure?](https://devenv.sh/guides/using-with-flakes/#getting-started)) in every terminal where you need the dev environment
+- Run `devenv up` in another terminal
+- Run `setup-devenv` in your terminal to install composer and npm packages and migrate and seed the database
+- Run `serve` to start serving the application on http://127.0.0.1:8000/
+
 ## Developing and contributing
 
 We want to let you know that Träwelling is a leisure project, developed and maintained by a team of volunteers who
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ If you want to use it:
 - [Install nix](https://github.com/DeterminateSystems/nix-installer) if you haven't already. (Make sure you've [enabled flakes and the nix command](https://nixos.wiki/wiki/Flakes#Permanent))
 - Activate the environment either by using the [direnv](https://direnv.net/) shell hook or by executing `nix develop --impure` ([why-impure?](https://devenv.sh/guides/using-with-flakes/#getting-started)) in every terminal where you need the dev environment
 - Run `devenv up` in another terminal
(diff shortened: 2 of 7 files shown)


flake.nix
```diff
@@ -0,0 +1,95 @@
...
+              echo "Initializing Passport"
+              ${php} artisan passport:install > /dev/null 2>&1
+            '';
+            serve.exec = ''
+              ${npm} run watch &
+              ${php} artisan serve
+            '';
+          };
+        };
+        formatter = pkgs.alejandra;
```
xanderio, line 91: I'm personally not all that keen on the way that alejandra formats nix code, but that more personal preference that anything else.

.gitignore
```diff
@@ -133,3 +133,7 @@ public/js/*.js.LICENSE.txt
 public/css/*.css
 public/css/*.css.map
 .phpunit.result.cache
+
+# Nix
```
xanderio, line 137: We could preemptively add `result*` to this section.

flake.nix
```diff
@@ -0,0 +1,95 @@
...
+  outputs = inputs @ {flake-parts, ...}:
+    flake-parts.lib.mkFlake {inherit inputs;} {
+      imports = [
+        inputs.devenv.flakeModule
+      ];
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
+        "x86_64-darwin"
+        "aarch64-darwin"
```
xanderio, line 17: Have you tested this flake on all these systems or is this just guess on which will probably work?

xanderio (approved):
Thanks, that looks really good. I've added some minor question to the PR, but non of them are blocking.

flake.nix
NyCodeGHG: I only tested on x86_64-linux, but I'm very sure it will work on all of the other platforms

.gitignore
NyCodeGHG: good idea

jeyemwey:
Does the flake.lock contain locks for the NPM and Composer packages? We'd need to move that out of the way to make dependabot etc work.

NyCodeGHG:
> Does the flake.lock contain locks for the NPM and Composer packages? 

No, flake.lock only locks the inputs of the flake.nix

Usually nix packages also use the native package managers's lockfile (but not here, because flake.nix only contains a devshell for now)

README.md
```diff
@@ -86,6 +86,23 @@ php artisan passport:install
...
+
+Nix is a cross-platform package manager for Linux and macOS systems.
+It also provides per project development environments.
+There is a also a Linux Distribution called NixOS which builds on top of nix,
+but it's not required to use nix and this development environment.
+
+Our [nix flake](flake.nix) includes such an environment with a pre configured MySQL instance.
+
+If you want to use it:
+
```
volkland, line 99: Previously, I relied heavily on Laravel's config manager to substitute the defaults and only wrote down the differences to the default in my `.env`. This doesn't really work if other applications also work with the `.env` file.

For newcomers, I'd like to point out that they need to copy the `.env` file.

```suggestion

- Copy the `.env.example` file to `.env`
```

I have later seen that the script does that for you on the first run, so __I'm open for debate__ if we should double the instruction here.

flake.nix
```diff
@@ -0,0 +1,95 @@
...
+            php.enable = true;
+            javascript.enable = true;
+          };
+          dotenv.disableHint = true;
+          services.mysql = {
+            enable = true;
+            ensureUsers = [
+              {
+                name = "homestead";
+                password = "secret";
```
volkland: This only works with the mysql information from `.env.example`. Can this please use the real environment variables? I'm not a flake/devenv expert, so I'm not sure how to handle that, but I'm sure its not a large task for you c:

flake.nix
```diff
@@ -0,0 +1,95 @@
...
+                echo "Migrating database"
+                ${php} artisan migrate:fresh --seed
+              else
+                echo "You seem to be not using mysql. Skipping migrations."
+              fi
+
+              echo "Generating Keys"
+              ${php} artisan key:generate > /dev/null 2>&1
+              echo "Initializing Passport"
+              ${php} artisan passport:install > /dev/null 2>&1
```
volkland: For some reasons, key:generate didn't work for me, but `2>&1` hid the error. Would have been nice to see the stderr.

flake.nix
volkland: Tested with `aarch64-darwin` and working like a charm.

volkland (changes requested):
Hi! Thanks for this great new feature. It really bumps up the developer experience for those using nix.

I have a few minor things that I'd like to improve, as seen in my comments. Nothing really show-stopping, just dotting some i's and crossing some t's.

When those are out of the way, my suggestion is to rebase the branch on top of develop and then squash it down to one single commit with the title `:technologist: add flake.nix development environment` before merging it upstream.

Edit: I'm just seeing that I wrote the review from my "work" account. This is @jeyemwey, btw :D

README.md
NyCodeGHG: a note that the nix environment does that automatically would be an idea maybe

flake.nix
NyCodeGHG: aarch64-linux works as well :)
träwelling running on an android phone

jeyemwey (approved):

PR #1864, opened 2023-08-21
Add Yearly Stats

wattnpapa:
This pull request adds yearly stats, styled similarly to the daily stats. 
We need to determine exactly where to integrate it.
For now, only views and controllers have been added.

There are no links to this feature from other pages at the moment.

I added this feature because I find it really nice to see where I've been within the current year and to track the routes I took during my travels.

### DIFF
--- a/app/Http/Controllers/Backend/Stats/YearlyStatsController.php
+++ b/app/Http/Controllers/Backend/Stats/YearlyStatsController.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace App\Http\Controllers\Backend\Stats;
+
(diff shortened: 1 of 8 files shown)


MrKrisKrisu:
Hi, thanks for your PR. :)
I'm honestly not sure if we should merge this. We have already tried a few times to offer yearly stats or bigger maps, but it is not possible with the amount of data. The browser would crash like this. This has also been seen well with the "old" enroute map. From about >~50 trips browser has no longer participated.

MrKrisKrisu:
Since no further discussion has taken place here for two weeks, I'll close.

PR #1866, opened 2023-08-22
Migrate to vite and es modules

NyCodeGHG:
This pr migrates from mix to the more modern [vite](https://vitejs.org) and the [Laravel Vite Plugin](https://laravel.com/docs/10.x/vite), and also migrates the usage of commonjs require calls to ecmascript modules.
There is stull some stuff in the js code which should be migrated to esm.

the nix stuff will be gone when #1857 is merged and this pr is rebased


NyCodeGHG:
should be ready for review, but merging is blocked by #1857

jeyemwey:
Hi, https://github.com/Traewelling/traewelling/commit/d1b98e08b3c82814c780d39822869cd4c732cc8e has fixed the lock-file issue which was blocking the unit tests. Can you please rebase your branch against upstream? Then the CI can run and this PR can be checked.

NyCodeGHG:
Seems like phpunit requires the vite manifest to be there?
I would prefer to make a production build in CI rather than commiting the manifest, since the mix manifest also only caused merge conflicts and it's more of a build artifact rather than source code

HerrLevin (approved):

PR #1867, opened 2023-08-22
:sparkles: Deliver operator for status via API

jheubuch:
With this PR, the HAFAS operator is delivered within the StatusResource.

### DIFF
--- a/app/Http/Resources/OperatorResource.php
+++ b/app/Http/Resources/OperatorResource.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace App\Http\Resources;
+
+use Illuminate\Http\Resources\Json\JsonResource;
+
+class OperatorResource extends JsonResource
+{
+    /**
+     * Transform the resource into an array.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return array|\Illuminate\Contracts\Support\Arrayable|\JsonSerializable
+     */
+    public function toArray($request)
+    {
+        return [
+            'id'    => $this->hafas_id,
+            'name'  => $this->name
+        ];
+    }
+}
--- a/app/Http/Resources/OperatorResource.php
+++ b/app/Http/Resources/OperatorResource.php
@@ -6,16 +6,10 @@ use Illuminate\Http\Resources\Json\JsonResource;
 
 class OperatorResource extends JsonResource
 {
-    /**
-     * Transform the resource into an array.
-     *
-     * @param  \Illuminate\Http\Request  $request
-     * @return array|\Illuminate\Contracts\Support\Arrayable|\JsonSerializable
-     */
     public function toArray($request)
     {
         return [
-            'id'    => $this->hafas_id,
+            'identifier'    => $this->hafas_id,
             'name'  => $this->name
         ];
     }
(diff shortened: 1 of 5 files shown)


app/Http/Resources/StatusResource.php
```diff
@@ -48,6 +48,10 @@ public function toArray($request): array {
                 'manualArrival'       => $this->trainCheckin->manual_arrival?->toIso8601String(),
                 'origin'              => new StopoverResource($this->trainCheckin->origin_stopover),
                 'destination'         => new StopoverResource($this->trainCheckin->destination_stopover),
+                'operator'            => [
```
HerrLevin: Please convert this to a nullable OperatorResource.

app/Virtual/Models/Train.php
```diff
@@ -142,4 +142,14 @@ class Train
...
      */
     private $destination;
+
+    /**
+     * @OA\Property (
+     *     title="operator",
+     *     description="Operator of the mean of transport"
+     * )
+     *
+     * @var Operator
```
HerrLevin, line 153: this needs to be nullable

HerrLevin (changes requested):

app/Http/Resources/StatusResource.php
jheubuch: How do I do that? 🙃 (I have no idea using PHP)

jheubuch: Ok, I tried and tested but maybe there is a better way

HerrLevin: You can create resources in Laravel with `php artisan make:resource OperatorResource`.

Then just have a look at the implementation of StopoverResource. :)

jheubuch: Ok I guess this is the better way than using `is_null(..) ? null : [...]` :D

HerrLevin: Exactly. :D

app/Http/Resources/OperatorResource.php
```diff
@@ -0,0 +1,22 @@
...
+     * Transform the resource into an array.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return array|\Illuminate\Contracts\Support\Arrayable|\JsonSerializable
+     */
+    public function toArray($request)
+    {
+        return [
+            'id'    => $this->hafas_id,
+            'name'  => $this->name
```
HerrLevin:
```suggestion
            'identifier'    => $this?->hafas_id,
            'name'  => $this?->name
```

app/Virtual/Models/Operator.php
```diff
@@ -0,0 +1,34 @@
...
+class Operator
+{
+    /**
+     * @OA\Property (
+     *     title="id",
+     *     example="sbb"
+     * )
+     * @var string
+     */
+    private $id;
```
HerrLevin:
```suggestion
    /**
     * @OA\Property (
     *     title="identifier",
     *     example="sbb"
     * )
     * @var string
     */
    private $identifier;
```

app/Http/Resources/StatusResource.php
```diff
@@ -48,6 +48,7 @@ public function toArray($request): array {
                 'manualArrival'       => $this->trainCheckin->manual_arrival?->toIso8601String(),
                 'origin'              => new StopoverResource($this->trainCheckin->origin_stopover),
                 'destination'         => new StopoverResource($this->trainCheckin->destination_stopover),
+                'operator'            => new OperatorResource($this?->trainCheckin->operator)
```
HerrLevin:
```suggestion
                'operator'            => $this->trainCheckin->HafasTrip->operator ? new OperatorResource($this->trainCheckin->HafasTrip->operator) : null
```

HerrLevin (changes requested):

PR #1887, opened 2023-09-02
Improvements for running tests etc. from flake

NyCodeGHG:
The Laravel test runner gets confused when the .env file is loaded in the environment.
The set environment override the .env.testing options, so they need to be unset for the tests to work correctly.
I added a small wrapper for artisan which also unsets the environment variables, so that part is done by laravel itself.
I also added a workaround for [this issue](https://github.com/cachix/devenv/issues/528#issuecomment-1556108767).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,7 +4,7 @@ MIX_APP_NAME=Träwelling
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
-APP_URL=http://localhost
+APP_URL=http://localhost:8000
 FORCE_HTTPS=false
 
 L5_SWAGGER_CONST_HOST=http://localhost:8000/api/v1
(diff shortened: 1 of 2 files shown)


jeyemwey (approved):
LGTM. Will merge soon :)

MrKrisKrisu:
> LGTM. Will merge soon :)

any update here?

jeyemwey:
Is the "unset" sandboxed in some way? Because what if the `artisan` command is called first, unsets the environment, and then `setup-devenv` or `devenv up` needs those variables?

NyCodeGHG:
unset doesn't modify the parent shell process, executing the script spawns a new bash process which doesn't modify its parent

jeyemwey:
> executing the script spawns a new bash process which doesn't modify its parent

Thanks, that was not clear to me before

PR #1888, opened 2023-09-02
Nix Package

NyCodeGHG:


NyCodeGHG:
i suggest that we register an organization on https://cachix.org to use their free tier for binary caching of our nix builds

NyCodeGHG:
So right now the package uses [composer2nix](https://github.com/svanderburg/composer2nix), but there's a promising PR in nixpkgs which adds a more modern composer builder which only requires to keep a hash up to date, just like the npm builder.
It seems to work pretty well with traewelling's codebase, i think we can use that as soon as it's merged.

See https://github.com/NyCodeGHG/traewelling/blob/6bf60ad064ca49f3e811d22cf6e0b6f1237f97f8/nix/package/default.nix
See https://github.com/NixOS/nixpkgs/pull/248184

NyCodeGHG:
~~I think this is ready as is for now~~
edit: nevermind

I'll also add the automated hash update thing in another PR, the old workflow in this PR was far too complicated and probably horrible to maintain

MrKrisKrisu:
I noticed this PR has been in a draft state for a while. Are you currently actively working on this? If not, would it be possible to close this PR for now and perhaps reopen it later when you're ready to proceed? This helps in maintaining a clearer overview of ongoing projects.

Thank you for your attention to this matter. 😊

NyCodeGHG:
i'm not actively working on it, we can close it for now

PR #1899, opened 2023-09-12
:stethoscope: Implement health monitoring routes

jeyemwey:
This PR introduces two health routes. The readiness probe (`/api/v1/health/ready`) checks if the database connection is available, and implicitly checks if the Laravel app is booted.
The second probe (`/api/v1/health/fitness`) does an more indepth check into the state of the application. It will:
* **check the database connection**
* **queue clog**. Basically: Are there suspiciously many jobs in the queue which makes us think that the queue process isn't working? From my understanding, each check-in spawns at least three jobs, plus webhooks, so I've put the threshhold to 9 Jobs waiting in the queue.
*  **db rest availability**. Does a real call to db-rest, just like the application would. This will break once FF/8000105 is removed for some reasons.
* **scheduler is up**. The PR sets up a `scheduler-canary` command which will set a cache value to the current `time()` once per minute. In the health route, it is checked when the canary the last time.

Currently missing:

- [ ] Update the `Dockerfile` to contain the `HEALTHCHECK` statement.
- [ ] Add an authentication to the route. _Debatable_, on the one hand, the routes reveal information over the instance which are potentially exploitable, on the other hand, apps may use the information and share it with their users. _Please discuss._
- [ ] If we decide to make the route public, we should also update our API documentation accordingly.
- [ ] After merge and rollout: Use the route in the status page and our internal monitoring. We can discuss this privately.

### DIFF
--- a/app/Http/Controllers/API/v1/HealthMonitorController.php
+++ b/app/Http/Controllers/API/v1/HealthMonitorController.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace App\Http\Controllers\API\v1;
+
+use App\Enum\CacheKey;
+use App\Http\Controllers\HafasController;
+use Illuminate\Database\QueryException;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Facades\Queue;
+
+class HealthMonitorController
+{
+
+    public function healthReady(): JsonResponse {
+        $checks = [
+            "database_reachable" => $this->isDatabaseReachable()
+        ];
+
+        return response()
+            ->json($checks)
+            ->setStatusCode(array_product($checks) ? 200 : 500);
+    }
+
+    public function healthFitness(): JsonResponse {
+        $checks = [
+            "database_reachable" => $this->isDatabaseReachable(),
+            "queue_running"      => $this->isQueueRunning(),
+            "db_rest_reachable"  => $this->isDbRestReachable(),
+            "scheduler_running"  => $this->isSchedulerRunning(),
+        ];
+
+        return response()
+            ->json($checks)
+            ->setStatusCode(array_product($checks) ? 200 : 500);
+    }
+
+    private function isDatabaseReachable(): bool {
+        try {
+            DB::connection()->getPdo();
+            return true;
+        } catch (QueryException $e) {
+            Log::info("Health Monitor ran into Query exception. Is the database not available?");
+            return false;
+        }
+    }
+
+    private function isQueueRunning(): bool {
+        $size = Queue::size();
+        return $size < 10;
+    }
+
+    private function isDbRestReachable(): bool {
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/API/v1/HealthMonitorController.php
```diff
@@ -0,0 +1,65 @@
...
+            return true;
+        } catch (QueryException $e) {
+            Log::info("Health Monitor ran into Query exception. Is the database not available?");
+            return false;
+        }
+    }
+
+    private function isQueueRunning(): bool {
+        $size = Queue::size();
+        return $size < 10;
```
MrKrisKrisu, line 52: This should be increased in my opinion, since our brouter tasks take a few seconds to retrieve. If many users check in parallel, the number increases quickly.

routes/api.php
```diff
@@ -43,6 +44,11 @@
     Route::get('static/privacy', [PrivacyPolicyController::class, 'getPrivacyPolicy'])
          ->name('api.v1.getPrivacyPolicy');
 
+    Route::get("/health/ready", [HealthMonitorController::class, 'healthReady'])
+         ->name("api.health.ready");
+    Route::get("/health/fitness", [HealthMonitorController::class, 'healthFitness'])
+         ->name("api.health.fitness");
```
MrKrisKrisu, line 50: Are you sure you want to make this information available to the public?

MrKrisKrisu (changes requested):

routes/api.php
jeyemwey: I think it can be useful for our app developers, but I'm open for a discussion.

app/Http/Controllers/API/v1/HealthMonitorController.php
jeyemwey: I think, I'd like to be conservative and keep the number quite low. When it starts to bug us, we can still update the check or set-up a second queue worker (the jobs are independent from each other and it should be easy to duplicate the systemd service, even on the old hardware).

routes/api.php
MrKrisKrisu: non blocking for me :)
I don't know any case where this can be abused, so...

jeyemwey:
Current State: After https://github.com/Traewelling/traewelling/pull/1910, I'm not sure if this code change is necessary.
In addition, I have a fear of duplicate code and stuff.

I'd move this PR to a draft unless someone else wants to work on this.

MrKrisKrisu:
I noticed this PR has been in a draft state for a while. Are you currently actively working on this? If not, would it be possible to close this PR for now and perhaps reopen it later when you're ready to proceed? This helps in maintaining a clearer overview of ongoing projects.

Thank you for your attention to this matter. 😊

jeyemwey:
I still like the idea of having a public health route, for our own "is trwl online" bot and for potential app developers. However, I cannot give a date when I can integrate the rest of the changes (see checklist). If anyone else feels compelled to do that, please do.

MrKrisKrisu:
Since there was no further interaction here, I'll close this PR. Feel free to open it up again when we start working on it again. :)

PR #1926, opened 2023-09-23
🔨 automatically generate Version after composer install

HerrLevin:


generate-version.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
NyCodeGHG:
```suggestion
#!/usr/bin/env bash
```
there are systems where `/bin/bash` does not exist

PR #1930, opened 2023-09-23
✨ Add parent_id to polyline, to make reverts easier

HerrLevin:


MrKrisKrisu:
This will create exceptions in the daily script which deletes unused polylines. Needs to be catched

PR #1947, opened 2023-09-30
:art: Move ".container"-element into Footer-Tag

tobiasreinhard:
For more consistency in the source code, the ".container" element outside the footer should be moved to the actual "footer"-tag.


tobiasreinhard:
Also fixed a minor layout issue with the color- and locale-switches. The content now uses the full size the column.

MrKrisKrisu (approved):
Thanks!

PR #1964, opened 2023-10-07
:ambulance: add timeout for most interactions with mastodon servers

jeyemwey:
This should cover most of the issues with bad-performing or offline mastodon servers by setting a timeout.

In a bad situation, it means that the mastodon-link in the profiles isn't there for an hour.

### DIFF
--- a/app/Http/Controllers/Backend/Social/MastodonProfileDetails.php
+++ b/app/Http/Controllers/Backend/Social/MastodonProfileDetails.php
@@ -23,7 +23,7 @@ class MastodonProfileDetails
     }
 
     private function getData(): ?array {
-        return Cache::remember('mastodon_'.$this->user->username, 6000, function () {
+        return Cache::remember("mastodon_{$this->user->username}", 60 * 60 /* 1 hour */, function() {
             return $this->fetchProfileInformation();
         });
     }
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Backend/Social/MastodonProfileDetails.php
```diff
@@ -23,7 +23,7 @@ public function getProfileUrl(): ?string {
     }
 
     private function getData(): ?array {
-        return Cache::remember('mastodon_'.$this->user->username, 6000, function () {
+        return Cache::remember("mastodon_{$this->user->username}", 60 * 60 /* 1 hour */, function() {
```
HerrLevin, line 26: NTH: i would prefer sprintf over double quote strings with variables but this also works.

app/Http/Controllers/Backend/Social/MastodonProfileDetails.php
```diff
@@ -34,12 +34,17 @@ private function fetchProfileInformation(): ?array {
...
+                                   ->call(
+                                       method:  "GET",
+                                       api:     "/accounts/" . $this->user->socialProfile->mastodon_id,
+                                       options: MastodonController::getRequestOptions()
+                                   );
                 }
             } catch (Exception $exception) {
                 // The connection might be broken, or the instance is down, or $user has removed the api rights
                 // but has not told us yet.
+                Log::warning("Unable to fetch mastodon information for user#{$this->user->id} for Mastodon-Server '{$mastodonServer->domain}' and mastodon_id#{$this->user->socialProfile->mastodon_id}");
```
HerrLevin, line 47: See above

HerrLevin (approved):

PR #1965, opened 2023-10-07
:bug: Only check for timezone offset, if the stop is not cancelled

HerrLevin:
Screenshot 2023-10-07 at 19 09 37

This only happened with cancelled trains, because cancelled trains are OBVIOUSLY UTC.

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -1,6 +1,8 @@
 <?php
 
+use Carbon\CarbonTimeZone;
 use Illuminate\Support\Carbon;
+use Illuminate\Support\Collection;
 
 /**
  * @see https://stackoverflow.com/a/437642
@@ -54,4 +56,16 @@ function userTime(null|Carbon|\Carbon\Carbon|string $time=null, ?string $format=
         return $time->tz($timezone)->isoFormat($format);
     }
     return $time->tz($timezone)->format($format);
+}
+
+function stationBoardTimezoneOffset(Collection $departures): bool {
+    foreach ($departures as $departure) {
+        if (!empty($departure?->cancelled) && $departure->cancelled) {
+            continue;
+        }
+        return \Carbon\Carbon::parse($departure->when)->tz->toOffsetName()
+               !== CarbonTimeZone::create(auth()->user()->timezone)->toOffsetName();
+    }
+
+    return false;
 }
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -1,5 +1,6 @@
 <?php
 
+use App\Models\User;
 use Carbon\CarbonTimeZone;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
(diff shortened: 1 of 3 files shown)


tests/Unit/HelperMethodTest.php
```diff
@@ -96,4 +99,46 @@ public static function userTimeDataProvider(): array {
...
+     * @dataProvider stationBoardTimezoneOffsetProvider
+     */
+    public function testStationBoardTimezoneOffset($expected, $departures): void {
+        $this->user = User::factory()->make();
+        Passport::actingAs($this->user);
+
+        $user = Mockery::mock($this->user)
+                       ->shouldReceive('getAttribute')
+                       ->with('timezone')
+                       ->andReturn('Europe/Berlin')
```
jeyemwey, line 112: Does that break in 3 weeks? `Europe/Berlin` does not explicitly say if its Daylight Savings Time, so I'm a little worried if the test will fail then.

tests/Unit/HelperMethodTest.php
```diff
@@ -96,4 +99,46 @@ public static function userTimeDataProvider(): array {
...
+        $wrong       = clone $correct;
+        $wrong->when = '2023-10-07T22:17:00+01:00';
+
+
+        return [
+            'cancelled, correct timezone'          => [false, [$cancelledCorrect, $correct]],
+            'cancelled, wrong timezone'            => [true, [$cancelledWrong, $wrong]],
+            'not cancelled, correct timezone'      => [false, [$correct, $cancelledCorrect]],
+            'not cancelled, wrong timezone'        => [true, [$wrong, $cancelledWrong]],
+            'cancelled, cancelled, wrong timezone' => [false, [$cancelledCorrect, $cancelledWrong]],
```
jeyemwey: What is the expected result when `$departures` is an empty array? Got a little lost here.

app/Helpers/Helper.php
```diff
@@ -55,3 +57,15 @@ function userTime(null|Carbon|\Carbon\Carbon|string $time=null, ?string $format=
     }
     return $time->tz($timezone)->format($format);
 }
+
+function stationBoardTimezoneOffset(Collection $departures): bool {
```
jeyemwey: Nitpick: Since the function returns a bool, I'd like to rename it to `hasStationBoardTimezoneOffsetToUser` and let it receive `auth()->user()` as a parameter. This also makes it easier to test, as you can just insert a minimal `App\Models\User` object instead of mocking stuff.

jeyemwey (changes requested):
Good contribution overall, few nitpicks and questions for my understanding.

tests/Unit/HelperMethodTest.php
HerrLevin: Oh, yes. You're definitely right.

I'll test it locally (+02:00 at that date **should** always compute to Europe/Berlin, but I'll verify it soon)

tests/Unit/HelperMethodTest.php
HerrLevin: False, technically. I'll add it.

tests/Unit/HelperMethodTest.php
HerrLevin: This comment saved me a lot of future headaches! Thanks you!

Technically, this still would've worked but only if you didn't search for the previous day (you're in CET and yesterday is in CEST) - now it'll always work.

Also added test cases for CET/CEST.

jeyemwey (approved):

PR #1986, opened 2023-10-15
🧑‍💻 reduce log severity in various places to improve log readability

HerrLevin:


app/Http/Controllers/Backend/BrouterController.php
```diff
@@ -91,7 +91,7 @@ public static function reroutePolyline(HafasTrip $trip): void {
             //2. Request route at brouter
             $brouterGeoJSON = self::getGeoJSONForRoute($coordinates);
         } catch (InvalidArgumentException) {
-            Log::error('[RefreshPolyline] Error while getting Polyline for HafasTrip#' . $trip->trip_id . ' (Required data is missing in Brouter response)');
+            Log::warning('[RefreshPolyline] Error while getting Polyline for HafasTrip#' . $trip->trip_id . ' (Required data is missing in Brouter response)');
```
MrKrisKrisu, line 94: I think error is correct here.

app/Models/TrainCheckin.php
```diff
@@ -75,7 +75,7 @@ public function getOriginStopoverAttribute(): TrainStopover {
                                                ->first();
         if ($stopOver == null) {
             //To support legacy data, where we don't save the stopovers in the stopovers table, yet.
-            Log::error('TrainCheckin #' . $this->id . ': Origin stopover not found. Created a new one.');
+            Log::info('TrainCheckin #' . $this->id . ': Origin stopover not found. Created a new one.');
```
MrKrisKrisu, line 78: This is also an error, as this should never happen...

app/Models/TrainCheckin.php
```diff
@@ -97,7 +97,7 @@ public function getDestinationStopoverAttribute(): TrainStopover {
                                                ->first();
         if ($stopOver == null) {
             //To support legacy data, where we don't save the stopovers in the stopovers table, yet.
-            Log::error('TrainCheckin #' . $this->id . ': Destination stopover not found. Created a new one.');
+            Log::info('TrainCheckin #' . $this->id . ': Destination stopover not found. Created a new one.');
```
MrKrisKrisu, line 100: This is also an error, as this should never happen...

MrKrisKrisu (changes requested):

MrKrisKrisu:
Maybe set the verbosity from the log lower instead?

app/Models/TrainCheckin.php
HerrLevin: This happens around 50 times a day and as long as we have checkins that were created before we introduced the new stopover format, this will still be the case. Error is (imho) a too high severity for a log message that is to be expected.

app/Models/TrainCheckin.php
HerrLevin: See above. It will always.

HerrLevin:
> Maybe set the verbosity from the log lower instead?

No. This pull request was created because I wanted to look for stack trace log occurrences in production, pointing to 500s. These have the log severity "error". Skipping through 10-20 useless "the origin stopover doesn't exist"-entries in my search results wasn't helping. :(

app/Http/Controllers/Backend/BrouterController.php
HerrLevin: Depends on how you're looking at it. This error is caused by outside dependencies. We cannot change the result of brouters response. If they can't find a route, I don't think that this should be an error because we can't fix it.

MrKrisKrisu (approved):

PR #2085, opened 2023-10-27
:bug: Always use the manual times if given; Use the time-is-real tooltip even if punctual

### LINKED ISSUE #2068: When the departure time is changed to the scheduled departure time, Träwelling displays the incorrect departure time.
**Describe the bug**
When a train is delayed after träwelling, you can't change the departure (real) time to the scheduled departure time.
For Example:
When I want to modify the departure time to 18:03 because the train was on time, träwelling ignores my input and continues to display the delay, such as when the train drives at 18:03 but after träwelling the train has a delay of 5 minutes, träwelling is showing the departure time of 18:08 completly ignoring the input.

**To Reproduce**

1. Checkin to a Train which has a delay
2. Edit the departure time to the scheduled departure time
3. See how Träwelling ignores the input
4. Click Edit and see that the input is saved correctly but gets ignored by träwelling

**Expected behavior**
Träwelling should show the time entered in the Departure (real) input if its the same as the time of the scheduled departure rather than keep showing the delay which isn't correct.

**Screenshots**

1. It shows the delay
Unbenannt

2. I try to edit the Departure (real)
Unbenannt

3. After Saving it still shows the wrong time.
Unbenannt

**Desktop (please complete the following information):**

- OS: Windows
- Browser: Firefox
- Version: 118.0.2

**Smartphone (please complete the following information):**

- Device: Huawei P30 Pro
- OS: Android 10
- Browser: Firefox
- Version: 119.0b9 #2015979707 d30a10afe7+

Also on Ipad with IpadOS 17 and newset Safari version

d3d9:
The manual time will always get used first, including the according tooltip, no matter whether it by chance matches with the planned time.
This fixes #2068.

This PR changes other behavior as well:
Previously, if there was realtime data, but no delay, the time-is-planned tooltip would be used. With this change, it will be distinguishable whether there is actually no delay according to the real-time data or whether there is no real-time data.

Here are a bunch of example screenshots.

---

A status with real-time data and delay.
rt-delay-no-edit

The departure time has manually been changed to the planned one, and the arrival was changed to a non-planned one (just as an example! When manually changing it to the planned one, it will also correctly use the manual time and tooltip -- previously, it would have used the "planned time" tooltip. The fact that the manual time matches the planned time can be recognized by there not being the stricken-through time).
rt-delay-edit-no-delay

The manual edits have been cleared:  
same as the first screenshot in this section.

---

A status with real-time data and no delay.
real-time-no-delay

The time has manually been changed to the planned one.
real-time-manual-no-delay

A delay was added manually.
real-time-manual-delay

The manual edits have been cleared:  
same as the first screenshot in this section.

---

A status without real-time data.
nort

A delay has manually been added.
nort-manual-delay

The time has manually been changed to the planned one (not cleared).
nort-manual-no-delay

The manual edits have been cleared:  
same as the first screenshot in this section.

### DIFF
--- a/app/Enum/TimeType.php
+++ b/app/Enum/TimeType.php
@@ -1,0 +1,23 @@
+<?php
+declare(strict_types=1);
+
+namespace App\Enum;
+
+enum TimeType: int
+{
+    case PLANNED = 0;
+    case REALTIME = 1;
+    case MANUAL = 9;
+
+    private function getTooltipStringId($timeType): string {
+        return match ($timeType) {
+            self::PLANNED  => 'time-is-planned',
+            self::REALTIME => 'time-is-real',
+            self::MANUAL  => 'time-is-manual'
+        };
+    }
+
+    public function getTooltip(): string {
+        return __(self::getTooltipStringId($this));
+    }
+}
--- a/app/Models/TrainCheckin.php
+++ b/app/Models/TrainCheckin.php
@@ -3,6 +3,7 @@
 namespace App\Models;
 
 use App\Casts\UTCDateTime;
+use App\Enum\TimeType;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
@@ -111,6 +112,47 @@ class TrainCheckin extends Model
             $this->HafasTrip->load('stopovers');
         }
         return $stopOver;
+    }
+
+    /**
+     * Takes the planned and optionally real and manual times and returns an array to use while displaying the status.
+     * Precedence: Manual > Real > Planned.
+     * Only returns the 'original' planned time if the updated time differs from the planned time.
+     */
+    private static function prepareDisplayTime($planned, $real = NULL, $manual = NULL): array {
+        if (isset($manual)) {
+            $time = $manual;
+            $type = TimeType::MANUAL;
+        } elseif (isset($real)) {
+            $time = $real;
+            $type = TimeType::REALTIME;
+        } else {
+            $time = $planned;
+            $type = TimeType::PLANNED;
+        }
+        return array(
+            'time' => $time,
+            'original' => ($planned->toString() !== $time->toString()) ? $planned : NULL,
+            'type' => $type
+        );
+    }
+
+    public function getDisplayDepartureAttribute(): \stdClass {
+        $planned = $this->origin_stopover?->departure_planned
+            ?? $this->origin_stopover?->departure
+            ?? $this->departure;
+        $real = $this->origin_stopover?->departure_real;
+        $manual = $this->manual_departure;
+        return (object) self::prepareDisplayTime($planned, $real, $manual);
+    }
+
+    public function getDisplayArrivalAttribute(): \stdClass {
+        $planned = $this->destination_stopover?->arrival_planned
+            ?? $this->destination_stopover?->arrival
+            ?? $this->arrival;
+        $real = $this->destination_stopover?->arrival_real;
+        $manual = $this->manual_arrival;
+        return (object) self::prepareDisplayTime($planned, $real, $manual);
     }
 
     /**
--- a/resources/views/includes/status.blade.php
+++ b/resources/views/includes/status.blade.php
@@ -41,27 +41,30 @@
                 <li>
                     <i class="trwl-bulletpoint" aria-hidden="true"></i>
                     <span class="text-trwl float-end">
-                        @if(isset($status->trainCheckin->manual_departure) && $status->trainCheckin->manual_departure->toString() !== $status->trainCheckin->origin_stopover->departure_planned->toString())
-                        <small style="text-decoration: line-through;" class="text-muted">
-                                {{ userTime($status->trainCheckin->origin_stopover->departure_planned) }}
-                            </small>
-                            &nbsp;
-                            <span data-mdb-toggle="tooltip" title="{{__('time-is-manual')}}">
-                                {{ userTime($status->trainCheckin->manual_departure) }}
-                            </span>
-                        @elseif($status->trainCheckin?->origin_stopover?->isDepartureDelayed)
+                        @php
+                            $original_departure = $status->trainCheckin->origin_stopover?->departure_planned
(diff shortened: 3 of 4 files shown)


jeyemwey:
Hi, thank you so much for your contribution! The idea with the tooltips is great to add to the accuracy of the statuses.

Since this is a quite complex construct (see the 10 screenshots that you had to add to explain the concept), I'd prefer if we added unit tests around it. Let me know if we can help you with that :)

d3d9:
Sure, would I do it in a similar way to [tests/Feature/Status/StatusBodyTest.php](https://github.com/Traewelling/traewelling/blob/develop/tests/Feature/Status/StatusBodyTest.php)?

I'd create a checkin, set the manual departure/arrival in different ways, then call the route and use some functions like assertSee(Text)(InOrder)? In the data provider, I would provide time offsets for by how much the manual edit should be done (null if there should not be a manual edit), separately for dep/arr, and whether the planned or real time should be used as the base of the offset-based modification, plus some boolean flags and string ids (for the tooltip content) to be used for checking with the various asserts.

It seems like I can't use CSS selectors unless there's browser testing or HTML parsing involved, so what would be the best way to go about testing e.g. the presence or absence of some response content separately for the departure and arrival (since the substrings like `<small style="text-decoration: line-through;" class="text-muted">` or the tooltip content look the same in both parts)?

Alternatively it might be better to go for another testing approach not based in the frontend, by moving the variable preparations out of the template? Not that the tests should be the main motivation for that, I just don't know whether somewhere else would have been a better place to add this logic to anyways?  
So basically we would prepare some properties that could be called e.g. `$display_departure`, `$display_departure_original` (could be null if there is no difference between the planned and displayed time, so no strikethrough time to be displayed), `$display_departure_type` (manual, real, planned).  
With my limited experience I have no idea about what's better, having some logic in the frontend template to the extent it currently has in my commit, or keep it out of there and do everything in something like a... controller? Looking at the FrontendStatusController, I'm struggling to understand whether / where the variables currently passed there are even used and whether the deprecation is relevant. I think I'll stop my research for now 😬

jeyemwey:
With what I know, I’d rather have it moved to a „real“ PHP function than in the templating engine. It’s just way easier to test and also reduces the complexity in the template. If needed, the template can change without having to touch the tests of a logic.

I would move the code to the TrainCheckin model as a public method (or an Laravel ˋgetDepartureDisplayAttributeˋ construct, as seen [here](https://github.com/Traewelling/traewelling/blob/4baec189df3753f59a6df11737b2ab337f2e1085/app/Models/TrainCheckin.php#L134)), since the business logic only works with data from that Model.

Instead of three properties, you can have a single method which returns an array or StdClass with the content. With the Type, I'd create an Enum in the ˋapp/enumsˋ folder. Ohhh, and you can deduplicate the code for arrival and departure with a „pure“ function that takes the three properties as arguments and doesn’t use the model object at all. I think that would reduce the complexity and redundant code even more.

> would I do it in a similar way to tests/Feature/Status/StatusBodyTest.php?

Thats a good idea, you can already see there how the TrainCheckin factory creates a basic checkin with a departure and arrival station and stuff. The factory definition is in ˋ database/factories/TrainCheckinFactory.phpˋ.

> I'm struggling to understand whether / where the variables currently passed there are even used

When you define the methods in the TrainCheckin model as described above, you can just access them in the template with ˋ$status->trainCheckin->departureDisplayˋ and Laravel will do some magic.

If you want to, you’re free to dabble with it, but don’t feel too pressed about perfectionism please. You already did a great job :)

d3d9:
Thank you for these useful pointers @jeyemwey. I have tried to improve my changes with the latest commit.

Does anybody have an idea why I had to use `@php` with `@endphp` instead of the one-line `@php(...)` here:
https://github.com/Traewelling/traewelling/blob/ac7bd368dc0fbe2a92ab0bd281f860cbc6d5126e/resources/views/includes/status.blade.php#L44-L46
while the following does not cause issues?
https://github.com/Traewelling/traewelling/blob/ac7bd368dc0fbe2a92ab0bd281f860cbc6d5126e/resources/views/includes/status.blade.php#L136

This is what I get when using the one-line variant in the first code block:
image

Just as an idea, this could maybe be extended to cover more fine-grained information based on the prognosisType from the db-rest response (prognosed, calculated, ...?) in the future.
https://github.com/Traewelling/traewelling/blob/ac7bd368dc0fbe2a92ab0bd281f860cbc6d5126e/app/Enum/TimeType.php#L8-L10

jeyemwey:
> Thank you for these useful pointers @jeyemwey. I have tried to improve my changes with the latest commit.

Looks good, thank you! There are a few things that I'd reorder before merging, but they're small and not pressing.

Did you have the option to look at unit tests? If not, I can add them myself :)

> Does anybody have an idea why I had to use `@php` with `@endphp` instead of the one-line `@php(...)` here:
> while the following does not cause issues?

I have no idea. Sometimes, php is just super weird.

> Just as an idea, this could maybe be extended to cover more fine-grained information based on the prognosisType from the db-rest response (prognosed, calculated, ...?) in the future.

Do you know if there's documentation to the prognosis type? I'm not sure what the information means and how/if we can display it to the user in a useful manner.

d3d9:
I just pushed my first attempt for tests. I basically only check manual edits based on the planned time.  
After glancing across the factories, I didn't find something related to real time data that would give me not just punctual, but also delayed trips, or even ones with no realtime data, since at first glance it seems like the realtime times are always set to the planned times?

> Do you know if there's documentation to the prognosis type? I'm not sure what the information means and how/if we can display it to the user in a useful manner.

I don't know details about these values. I assume these are not the most interesting ones that could theoretically exist. It would be very interesting though if, in the future, there would be information about whether it's already "confirmed" real-time data and no prognosis anymore, the same way e. g. bahn.expert shows the time in bold once a train has actually departed, based on the door closing or whatever. Since we're probably far away from having this data in the HAFAS responses, this idea is not relevant right now, I guess.

jeyemwey:
I updated the test, reducing the complexity of the test case by not testing everything at once. Hope you like it and can follow what i did.

At this point, im a little too involved, and I'd like to have @MrKrisKrisu's opinion on the issue, if he has time c:

MrKrisKrisu (approved):
Sorry for the late review - thank you! :)
works fine for me

PR #2096, opened 2023-10-31
feat: remove unnecessary cookie alert

pReya:
- remove the cookie notice/alert since it's legally not required

According to GDPR, a cookie banner is only necessary for cookies which are NOT technically required for the website to work properly. Since the only cookie that Traewelling is setting is a default Laravel session cookie, it is absolutely "technically required". There is no need to ask for consent or even show a cookie notice at all for this cookie. Mentioning it in the "Privacy" page is already enough.

My partner is a privacy lawyer, so we talk about this stuff a lot, and Traewelling is a pretty clear case ;)


pReya:
The imports from the `bootstrap-cookie-alert` package: Where are they defined? I couldn't find anything in the `package.json` – so how can I remove this package?

MrKrisKrisu:
I've thought about it before, but discarded the idea when I took another closer look at Träwelling. I haven't really looked into the legal perspective of data protection at Träwelling, @HerrLevin has taken care of it so far.

But when I look at the Träwelling landing page as a non-logged in user, I receive two cookies:

1. `XSRF-TOKEN` -> technically necessary
2. `trwl_session` -> I'm unsure here.

The cookie stores a unique ID of the visitor - and then also in the database. This seems to come from a standard Laravel (UI?)  implementation, but I don't know why this is necessary. I would like to clarify that beforehand.

image

pReya:
The session cookie is there to identify the user, which is the main (one might even say the only) part of Träwellings functionality. It is (without a doubt) technically necessary. You will find a lot of consistent information on this topic if you google "session cookie gdpr".

The relevant section of the law in this regard is § 25 (2) Nr. 2 TTDSG (https://dsgvo-gesetz.de/ttdsg/25-ttdsg/)

NyCodeGHG:
There's no reason to identify a user if they are not logged in

pReya:
> There's no reason to identify a user if they are not logged in

I agree, the session cookie is currently already being set, when the home page is loaded – this is unnecessary. But it's also a different ticket/issue in my opinion, and not related to the cookie alert.

EDIT: I've created a new issue for this behavior: https://github.com/Traewelling/traewelling/issues/2101

MrKrisKrisu:
I've added the label `needs discussion` because I think we need to first discuss about #2101 before rolling this out.

I totally agree that we don't want any cookies which need the agreement of the user - and I understand that the login cookie is technically required.

But as already mentioned, laravel is saving a default session cookie for every user - including the ip and user agent. In my opinion this should be fixed first.

pReya:
> I've added the label `needs discussion` because I think we need to first discuss about #2101 before rolling this out.
> 
> I totally agree that we don't want any cookies which need the agreement of the user - and I understand that the login cookie is technically required.
> 
> But as already mentioned, laravel is saving a default session cookie for every user - including the ip and user agent. In my opinion this should be fixed first.

I understand your position. But legally, these are totally separate topics. The cookie notice is not required – even if #2101 is not implemented / implemented at a later point in time.

MrKrisKrisu:
Since there was no further interaction or other opinions here, I would close the PR for now.
If issue #2101 is resolved, I would welcome the removal of the cookie banner!

pReya:
> Since there was no further interaction or other opinions here, I would close the PR for now. If issue #2101 is resolved, I would welcome the removal of the cookie banner!

Just to repeat myself: These are unrelated topics. This can be merged, even without the other issue.

PR #2098, opened 2023-11-05
:sparkles: add status tags to ui (view only)

MrKrisKrisu:
Anyone some font-awesome icon ideas?

Screenshot 2023-11-05 at 09 58 59

### DIFF
--- a/resources/views/status.blade.php
+++ b/resources/views/status.blade.php
@@ -46,20 +46,8 @@
                 @foreach($status->tags as $tag)
                     @can('view', $tag)
                         <span class="badge bg-trwl">
-                            @if($tag->key === 'trwl:seat')
-                                <i class="fa-solid fa-couch me-1"></i>
-                            @elseif($tag->key === 'trwl:ticket')
-                                <i class="fa-solid fa-qrcode me-1"></i>
-                            @elseif($tag->key === 'trwl:wagon')
-                                <!-- TODO: Icon? -->
-                            @elseif($tag->key === 'trwl:travel_class')
-                                <!-- TODO: Icon? -->
-                            @elseif($tag->key === 'trwl:locomotive_class')
-                                <!-- TODO: Icon? -->
-                            @elseif($tag->key === 'trwl:wagon_class')
-                                <!-- TODO: Icon? -->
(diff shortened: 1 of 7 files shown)


vainamov:
- `<i class="fa-solid fa-ticket-simple"></i>` for the ticket type?
- `<i class="fa-solid fa-left-right"></i>` for the carriage number?
- `<i class="fa-solid fa-couch"></i>` instead of chair for the seat?

resources/views/status.blade.php
```diff
@@ -42,6 +42,29 @@
                         {{$status->trainCheckin->HafasTrip->last_refreshed->diffForHumans()}}
                     </small>
                 @endif
+
+                @foreach($status->tags as $tag)
+                    @can('view', $tag)
+                        <span class="badge bg-trwl">
+                            @if($tag->key === 'trwl:seat')
```
HerrLevin: Maybe add an Enum with `\App\Interfaces\IconEnumInterface` to keep all icons at one central point and to reduce clutter?

MrKrisKrisu:
> * `<i class="fa-solid fa-ticket-simple"></i>` for the ticket type?

Hmm, I'm not sure - I like the QR code for the ticket.

> * `<i class="fa-solid fa-left-right"></i>` for the carriage number?

I don't understand, how arrows represents the carriage number. :c 

> * `<i class="fa-solid fa-couch"></i>` instead of chair for the seat?

I like the couch instead of the seat, thanks. :)

vainamov:
> Hmm, I'm not sure - I like the QR code for the ticket.

Fair enough :D

> I don't understand, how arrows represents the carriage number. :c

Now that I think about it, I wouldn't either. When I chose it, I thought it ever-so-slightly resembled a train from the side.

But considering the style of the tags, I don't think icons are necessary for each of them. So I would rather spare an icon than pick a less-suitable one.

HerrLevin:
Maybe `<i class="fa-solid fa-van-shuttle"></i>` for the carriage number?
`<i class="fa-solid fa-ticket"></i>` for the class?

HerrLevin (approved):

PR #2115, opened 2023-11-12
📝 Correct success response on notifications

### LINKED ISSUE #2108: API: Wrong response of `PUT notifications/read/all`
API docs suggest a `message` attribute as response but in reality this `message` attribute is nested into a `data` attribute.

HerrLevin:
Fixes #2108

Screenshot 2023-11-12 at 13 29 37


marhei:
So you really want to establish a new type of response? In my opinion, it makes no sense that `/notifications/read/all` returns a message in data response, while `settings/profilePicture` returns a pure message response, for example.

marhei:
Nevermind, the `settings/profilePicture` endpoint also has the message attribute inside a data attribute…

HerrLevin:
@marhei I mean... the profilePicture-endpoint is not yet documented. We could just return an empty 200?

marhei:
I would propose the same return as the `DELETE /webhooks/{id}` endpoint with a unlocalized `success` status. 

 image

app/Http/Controllers/API/v1/NotificationsController.php
```diff
@@ -176,6 +172,6 @@ public function markAsUnread(string $notificationId): JsonResponse {
      */
     public function markAllAsRead(): JsonResponse {
         Auth::user()->unreadNotifications->markAsRead();
-        return $this->sendResponse(['message' => __('notifications.readAll.success')]);
+        return $this->sendResponse();
```
MrKrisKrisu, line 175: This may break our frontend. Please double check.

HerrLevin: Screenshot 2023-11-12 at 14 37 44
oops

resources/js/api/api.js
```diff
@@ -27,7 +27,9 @@ export default class API {
         }
 
         return response.json().then(data => {
-            notyf.success(data.data.message);
+            if (data && data.data && data.data.message) {
```
MrKrisKrisu, line 30: So now there is just no notification for `read all notifications`?

HerrLevin: With this solution: Yes.

We'll have to add translations to vue since it's bad practice to have the API return "static" localized strings. It would be better to implement this in the frontend but I don't know which i18n package to use yet. 

Either we prioritize good api design and ignore the notification for now (probably not a bad idea since the list automatically clears on success. imho there's no notification needed), or we wait until we have i18n sorted out and draft this PR.

HerrLevin:
Only merge after #2177 2177

MrKrisKrisu (approved):

PR #2152, opened 2023-11-21
:children_crossing: Improve recognition of data as realtime

d3d9:
Currently, any incoming realtime data for past stopovers is ignored while any future time value is interpreted as realtime data.  
In the following places, the assumption is made that the presence of the `arrival` and `departure` attributes in the db-rest FPTF response implies these values being realtime data or that any time in the past can never be realtime data.

https://github.com/Traewelling/traewelling/blob/4baec189df3753f59a6df11737b2ab337f2e1085/app/Http/Controllers/HafasController.php#L445-L456

https://github.com/Traewelling/traewelling/blob/4baec189df3753f59a6df11737b2ab337f2e1085/app/Http/Controllers/HafasController.php#L499-L501

https://github.com/Traewelling/traewelling/blob/4baec189df3753f59a6df11737b2ab337f2e1085/app/Http/Controllers/HafasController.php#L483-L487

Looking at the actual responses though, realtime data for stopovers up to 15 minutes in the past could be included in many cases. The way to recognize this is using the `arrivalDelay` and `departureDelay` attributes, which are `null` if it isn't realtime data, and `0` or another number if it is.  
Here's an example of such a response: https://gist.github.com/d3d9/7b9c494a4ccb3d44618dd885b656cfcc.  
It was made at 16:18 and included realtime data for the stopovers starting from Mercedes Jürgens, which had a realtime departure of 16:04 and planned departure of 16:02.

The following screenshots show some of the improvements, which could partially even be described as bug fixes.
Sorry for not matching the language in the before/after screenshots :p

These trip requests were made at the same time, 23:05, without (left) or with the changes from this PR (right).

These are trip requests for the same trip a month into the future. Currently, the +0 gets displayed since any data from the future is handled as if it's realtime..

This is a checkin made 5 minutes after the realtime arrival, and the realtime arrival is saved correctly. I didn't create a side-by-side screenshot here, but I tested it concurrently on another instance without the changes and there, only the departure was realtime (sourced from the stationboard I assume), while the arrival did not have realtime data.

___

Other notes:

https://github.com/Traewelling/traewelling/blob/4b9dad73da858c0ef3906733306ec77b07172fd4/app/Http/Controllers/HafasController.php#L485

Assumption here: e. g. arrivalDelay could be null for the first stopover which is perfectly fine, but there shouldn't be a situation where e. g. arrivalDelay is null for a non-first stopover while departureDelay is **not** null for that same stopover. As long as that's true, there should be no need to worry about, in this example, previously saved realtime arrival data being lost during a refresh.  
If we want to be more precise, this should be split up in a way only the arrival or the departure get updated in case one of the respective attributes is null to avoid overriding previously saved realtime data for either arrival or departure.

### DIFF
--- a/app/Http/Controllers/HafasController.php
+++ b/app/Http/Controllers/HafasController.php
@@ -442,13 +442,15 @@ abstract class HafasController extends Controller
             //remove "null" values
             $updatePayload = array_filter($updatePayload, 'strlen'); //TODO: This is deprecated, find a better way
 
-            if ($stopover->arrival !== null && Carbon::parse($stopover->arrival)->isFuture()) {
+            //the arrival and departure attributes are always included, so to recognize whether we have realtime data,
+            // arrivalDelay and departureDelay are checked for being null or not.
+            if ($stopover->arrival !== null && isset($stopover->arrivalDelay)) {
                 $updatePayload['arrival_real'] = Carbon::parse($stopover->arrival);
                 if ($stopover->arrivalPlatform !== null) {
                     $updatePayload['arrival_platform_real'] = $stopover->arrivalPlatform;
                 }
             }
-            if ($stopover->departure !== null && Carbon::parse($stopover->departure)->isFuture()) {
+            if ($stopover->departure !== null && isset($stopover->departureDelay)) {
                 $updatePayload['departure_real'] = Carbon::parse($stopover->departure);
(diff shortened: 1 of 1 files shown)


MrKrisKrisu (approved):
workes fine - thanks!

PR #2153, opened 2023-11-22
:sparkles: add WebFinger support

NyCodeGHG:


routes/web.php
```diff
@@ -35,6 +35,7 @@
 use App\Http\Controllers\PrivacyAgreementController;
 use App\Http\Controllers\SitemapController;
 use App\Http\Controllers\UserController;
+use App\Http\Controllers\Backend\WebFingerController;
```
MrKrisKrisu: Please don't use Controllers from the backend here.

MrKrisKrisu (approved):

PR #2177, opened 2023-12-01
🌐  add frontend translations to vue components

HerrLevin:


HerrLevin:
Merge #2115 after this

resources/js/app.js
```diff
@@ -42,24 +43,49 @@ window.notyf = new Notyf({
 });
 
 document.addEventListener("DOMContentLoaded", function() {
+    // get language query parameter
+    let fallbackLang = "en";
+    const urlParams = new URLSearchParams(window.location.search);
```
NyCodeGHG, line 48: i don't think storing the language in the url params is a good idea.
it's really annoying when you get a link from someone who uses a different language than you and you have to manually fix it to use the language you want to see

HerrLevin: We're not storing the language in the url params. That's the current blade-implementation of how we set it. The idea here is to once set the fallback in the global i18n instance and then forget about it until it's set again. (With the current blade/vue-setup this won't always work until we've migrated to vue - but it's at least something.

MrKrisKrisu (approved):

PR #2207, opened 2023-12-12
:sparkles: add automatic account deletion

MrKrisKrisu:


MrKrisKrisu:
Due to the risk of this code I would like to have a minimum of **two** approves instead of the required one.

Any external reviewers are welcome too!

MrKrisKrisu:
I would merge and deploy this in the first week of Jan 2024, if no one has something against it.

PR #2220, opened 2023-12-18
add logic to model

MrKrisKrisu:
Currently we have origin, destination, departure and arrival in the train_checkins table. This is not well-designed, because we have to store the same information multiple times. This migration adds two new columns for a foreign key to the stopovers table, so we can remove the duplicate information in the future.

This is intentionally kept so small to make migration easier. The double columns should be gradually changed in the code.

Also added a migration command - I think we should first migrate the data before adapt the code. This is much easier.


MrKrisKrisu:
ah shit, wrong branch selected

PR #2229, opened 2023-12-21
:monocle_face: improve caching of checkin duration

MrKrisKrisu:


app/Observers/CheckinObserver.php
```diff
@@ -2,16 +2,16 @@
...
 namespace App\Observers;
 
+use App\Http\Controllers\Backend\Transport\TrainCheckinController;
 use App\Models\TrainCheckin;
 
 class CheckinObserver
 {
 
     public function updated(TrainCheckin $checkin): void {
         if ($checkin->isDirty(['origin', 'destination', 'departure', 'arrival'])) {
```
d3d9: Would it make sense to add `manual_departure` and `manual_arrival` here?

MrKrisKrisu: oh, well. yes. thanks!

PR #2251, opened 2023-12-29
:sparkles: save client id if status is created via app

MrKrisKrisu:


app/Http/Resources/StatusResource.php
```diff
@@ -30,23 +30,26 @@ public function toArray($request): array {
             'likes'          => (int) $this->likes->count(),
             'liked'          => (bool) $this->favorited,
             'isLikable'      => Gate::allows('like', $this->resource),
+            'client'         => [
```
HerrLevin: I would prefer a ClientResource, so that all api consumers will get the name and url of an app

app/Http/Resources/StatusResource.php
```diff
@@ -30,23 +30,26 @@ public function toArray($request): array {
...
+                'id' => $this->client_id,
```
HerrLevin: Also please update the api docs

HerrLevin (changes requested):

app/Http/Resources/ClientResource.php
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Model -> OAuthClient
+ */
+class ClientResource extends JsonResource
+{
+    public function toArray($request) {
+        return [
+            'id'   => $this->id,
+            'name' => $this->name,
```
HerrLevin: Maybe add URL (+ ToS/Privacy URL)?

app/Virtual/Models/Client.php
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @OA\Schema(
+ *     title="Client",
+ *     description="Client model",
+ *     @OA\Xml(
+ *         name="Client"
+ *     )
+ * )
+ */
+class Client
```
HerrLevin, line 14: See model

resources/views/admin/status/edit.blade.php
```diff
@@ -34,11 +34,21 @@
...
                                 {{ $status->checkin->trip_id }}
                             </a>
                         </div>
+
+                        <div class="col-4">
+                            <label>Client</label>
+                        </div>
+                        <div class="col-8">
+                            @isset($status?->client)
+                                {{$status->client->name}} (#{{$status->client->id}})
```
HerrLevin, line 48: URL/Link

HerrLevin (changes requested):

app/Http/Resources/ClientResource.php
MrKrisKrisu: Isn't this a little bit too much for every GET request on statuses?

resources/views/admin/status/edit.blade.php
MrKrisKrisu: What should be linked here in the admin panel?

app/Virtual/Models/Client.php
MrKrisKrisu: which model should I see at?

app/Http/Resources/ClientResource.php
HerrLevin: No I don't think so. A few strings won't hurt. :D
We send way more information with every event, train, etc. It's more efficient to send the url(s) for the client, rather than have another client request the urls separately.

app/Virtual/Models/Client.php
HerrLevin: I mean resource 🫣

resources/views/admin/status/edit.blade.php
HerrLevin: Nvm. Didn't see the admin panel.

HerrLevin (approved):

PR #2253, opened 2023-12-31
:bug: rename classname in test

MrKrisKrisu:


jeyemwey:
Was the test previously broken because of the broken import?

MrKrisKrisu:
> Was the test previously broken because of the broken import?

Yes, I didn't merge the current develop (with changes from #2234) into #2207, so the old classname was merged there.

PR #2304, opened 2024-01-29
:children_crossing: improve events form in backend

MrKrisKrisu:


MrKrisKrisu:
merged without review because this affects only the backend.

PR #2322, opened 2024-02-06
:bug: Fix "uneditable" 0min checkins & allow checking in to a destination 0 min away

d3d9:
It is currently not possible to edit a checkin where the arrival at the destination is the same time as the departure from the origin without changing the destination. You are forced to change the destination or to delete and recreate the checkin.

Before:

After:

---

I was wondering whether there are more issues related to these lovely 0 minute durations between stopovers and I found another issue here:

Trip view for start at stopover Obernahmer: it is possible to check in with destination Obernahmer Land or Lahmen Hasen.

Trip view for start at stopover Obernahmer Land: it is not possible to check in towards Lahmen Hasen.

With the fix applied 🐇

---

I didn't deduplicate the code because I kept using the stopover for the existing checkin edit case, and station + departure for the new trip case. I also don't use the planned arrival anymore inside getAlternativeDestinationsForCheckin since the remaining code determines the exact stopover after which every remaining stopover will pass.
I hope there are no cases where the condition never evaluates to true; or where the pair of (departure + stopover), or (departure + station) are not uniquely identifying exactly one stopover of a trip.

### DIFF
--- a/app/Http/Controllers/Backend/Transport/StationController.php
+++ b/app/Http/Controllers/Backend/Transport/StationController.php
@@ -69,9 +69,14 @@ abstract class StationController extends Controller
     }
 
     public static function getAlternativeDestinationsForCheckin(Checkin $checkin): Collection {
+        $encounteredOrigin = false;
         return $checkin->trip->stopovers
-            ->filter(function(Stopover $stopover) use ($checkin) {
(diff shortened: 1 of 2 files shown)


MrKrisKrisu (approved):
oh, well. :D Thanks.

PR #2326, opened 2024-02-07
:bug: Fix selection of trips with active checkins to be refreshed

d3d9:
With the previous logic, even trips days away were refreshed regularly, making lots of unnecessary requests and potentially blocking more useful ones if the defined limit of trips to be refreshed were to be reached.

My first attempt was still based on each stopover like the previous logic:

```diff
   ->where(function($query) {
            $query->where('train_stopovers.arrival_planned', '>=', now()->subMinutes(20))
                  ->orWhere('train_stopovers.arrival_real', '>=', now()->subMinutes(20));
        })
   ->where(function($query) {
            $query->where('train_stopovers.departure_planned', '<=', now()->addMinutes(20))
                  ->orWhere('train_stopovers.departure_real', '<=', now()->addMinutes(20));
        })
```

However, imagine you have a train that doesn't have any stopovers for an hour, in that case there would be no refreshes during a third of that hour.

That's why I changed it the way you can see in this PR branch, using the origin and destination stopovers of the checkin and not joining every stopover anymore.

I think it would good to have a padding around the checkin, since critical realtime updates could still come in even when Träwelling thinks it's over, and to keep showing fresh data ahead of the checkin as well, so I added 20 minutes before and after as an example to be discussed.

### DIFF
--- a/app/Console/Commands/RefreshCurrentTrips.php
+++ b/app/Console/Commands/RefreshCurrentTrips.php
@@ -16,16 +16,19 @@ class RefreshCurrentTrips extends Command
     protected $description = 'Refresh delay data from current active trips';
 
     public function handle(): int {
-        $this->info('Gettings trips to be refreshed...');
+        $this->info('Getting trips to be refreshed...');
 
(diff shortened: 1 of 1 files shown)


MrKrisKrisu (approved):
Thanks! :)

PR #2328, opened 2024-02-07
:sparkles: Remove webhook from database when webhook client has responded with 410 Gone

jeyemwey:
Background: Some users have a large number of webhooks configured, often for the same URL, i.e. because they uninstalled and reinstalled an app. The old webhook will usually not be discarded, so when webhook events are fired, dozens of requests are sent out. The webhook client (FCM, APNS, ...) knows if the app is still installed and can respond to us if it isn't anymore.

**TODO:** I have added unit tests as far as I understood the library to work. However, I don't have a working webhook-able installation here, so if @MrKrisKrisu or @NyCodeGHG could check out the feature on their side, that would be great.

### DIFF
--- a/app/Listeners/RemoveAbsentWebhooksListener.php
+++ b/app/Listeners/RemoveAbsentWebhooksListener.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace App\Listeners;
+
+use App\Enum\CacheKey;
+use App\Enum\MonitoringCounter;
+use App\Models\Webhook;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\Log;
+use Spatie\WebhookServer\Events\WebhookCallFailedEvent;
+
+class RemoveAbsentWebhooksListener
+{
+    public function handle(WebhookCallFailedEvent $event) {
+        if (!$event->response) {
+            return;
+        }
+
+        if ($event->response->getStatusCode() !== 410) {
+            return;
+        }
+
+        $webhookId = $event->headers["X-Trwl-Webhook-Id"];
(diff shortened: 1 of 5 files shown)


MrKrisKrisu:
> **TODO:** I have added unit tests as far as I understood the library to work. However, I don't have a working webhook-able installation here, so if @MrKrisKrisu or @NyCodeGHG could check out the feature on their side, that would be great.

To be honest, I never worked with our webhooks implementation.

app/Listeners/RemoveAbsentWebhooksListener.php
```diff
@@ -0,0 +1,31 @@
...
+class RemoveAbsentWebhooksListener
+{
+    public function handle(WebhookCallFailedEvent $event) {
+        if (!$event->response) {
+            return;
+        }
+
+        if ($event->response->getStatusCode() !== 410) {
+            return;
+        }
```
MrKrisKrisu:
```suggestion
        if (!$event->response || $event->response->getStatusCode() !== 410) {
            return;
        }
```

HerrLevin:
@jeyemwey You could test it with [traewelling-telegram-notifications](https://github.com/HerrLevin/traewelling-telegram-notifications). I'll also test it tomorrow if I find the time.

NyCodeGHG:
I primarily implemented webhooks for träwelldroid being able to implement push notifications, so I don't really have something to test this quickly with either.

But code looks good to me

HerrLevin:
Works on my end! :)

MrKrisKrisu (approved):
Code and tests are looking good. :)
Didn't test locally.

PR #2331, opened 2024-02-07
:children_crossing: show profile picture current map position

MrKrisKrisu:
image

image

### DIFF
--- a/app/Dto/LivePointDto.php
+++ b/app/Dto/LivePointDto.php
@@ -3,7 +3,10 @@
 namespace App\Dto;
 
 use App\Dto\GeoJson\Feature;
-use OpenApi\Annotations as OA;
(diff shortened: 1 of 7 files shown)


app/Dto/LivePointDto.php
```diff
@@ -97,7 +105,17 @@ public function toArray(): array {
             'arrival'   => $this->arrival,
             'departure' => $this->departure,
             'lineName'  => $this->lineName,
-            'statusId'  => $this->statusId
+            'statusId'  => $this->status->id, //deprecate this... why aren't we using a resource here...? why is a DTO a resource now? isn't a DTO supposed to be a simple data transfer object for internal use?
+            'status'    => [
```
HerrLevin, line 109: since this is a temporary solution, you could remove this and add a prop to the vue element.

HerrLevin (approved):

PR #2374, opened 2024-02-19
:sparkles: add Co2 savings to profile view

tomizius:
I want to see how much co2 I've safed against driving by car with the train during my live.
I used following source https://www.quarks.de/umwelt/klimawandel/co2-rechner-fuer-auto-flugzeug-und-co/

Unfortunately it's not easy to calculate how much the user has been driven by ÖPNV or IC(E). So I used the factor 15 to calculate the value and divided it by 100.

### DIFF
--- a/resources/views/profile.blade.php
+++ b/resources/views/profile.blade.php
@@ -43,6 +43,9 @@
                     <span class="profile-stats">
                             <span class="font-weight-bold"><i class="fa fa-route d-inline"></i>&nbsp;{{ number($user->train_distance / 1000) }}</span><span
                                 class="small font-weight-lighter">km</span>
+			    <span class="font-weight-bold ps-sm-2">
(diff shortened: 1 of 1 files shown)


HerrLevin (approved):

MrKrisKrisu:
I am unsure whether I like this PR. On the one hand, I find an attribute more in the header of the profile is too much - especially in the mobile view. I also find the calculation of the saved CO2 too flat. The differences between bus, train, car, ... are not taken into account enough.

image

MrKrisKrisu (changes requested):
I am unsure whether I like this PR. On the one hand, I find an attribute more in the header of the profile is too much - especially in the mobile view. I also find the calculation of the saved CO2 too flat. The differences between bus, train, car, ... are not taken into account enough.

image

MrKrisKrisu:
belongs to discussion #2327

MrKrisKrisu:
As there was no further interaction, I refer to discussion #2327 and close the PR for the time being.

PR #2398, opened 2024-03-07
🐛 Fix visibility dropdown ui

HerrLevin:


MrKrisKrisu (approved):
I don't understand Vue, but looks okay.
giphy

PR #2407, opened 2024-03-10
🐛 Fix issues related to checking into international trains

### LINKED ISSUE #2401: Intermediate stops not appearing
### Describe the bug

For the last month or so, Traewelling has for some reason not shown any intermediate stops when entering a journey:

IMG_7606

I can enter a journey to the final stop using the “fast forward” icon, but that’s it. 

### Steps to reproduce

Try to enter any journey on a UK train, as far as I can tell; I’ve mostly been doing journeys from Leuchars(St.Andrews)

### Browser console logs

_No response_

### Browser

iOS Safari

fabiancdng:
As issue #2401 already mentioned, checking into UK trains does not seem to work. Neither using the old check-in interface nor using the new (experimental) one.

I've never worked on the Träwelling codebase before, so take this with a grain of salt 😅

## New check-in interface
To me it looks like the issue with the new interface was due to many UK trains not having a line name provided in the data. Some Vue components were not accounting for that.
I fixed this by checking for empty line names and using the `fahrtNr` as a fallback. This seemed to be the preferred solution looking at the old interface.
This actually seems to resolve the display glitches and the 422 errors when trying to check into a train:
Screenshot 2024-03-10 at 18 49 50

## Old check-in interface
However, the old interface doesn't seem to have any issues with empty line names. The only problem there is the empty stopovers. As far as I can tell, this issue is related to the filtering of stopovers introduced in #2322.

The UK trains seem to provide the planned departure (`departure_planned`) in their own timezone. Hence, checking against the departure time passed by the front end does not work (as the UK time is one hour behind).

I removed the departure time check and only kept the check against the start station passed by the front end. I hope this is sufficient? Not entirely sure on that one though.

Before:
Screenshot 2024-03-10 at 18 47 15

After:
Screenshot 2024-03-10 at 18 48 04

Resolves #2401

### DIFF
--- a/app/Http/Controllers/FrontendTransportController.php
+++ b/app/Http/Controllers/FrontendTransportController.php
@@ -127,6 +127,7 @@ class FrontendTransportController extends Controller
                 // in long term to support multiple data providers we only support IDs here - no IBNRs.
                 $startStation = Station::findOrFail($validated['start']);
             }
+
             $departure = Carbon::parse($validated['departure']);
 
             $trip = TrainCheckinController::getHafasTrip(
@@ -137,9 +138,9 @@ class FrontendTransportController extends Controller
 
             $encounteredStart = false;
             $stopovers = $trip->stopovers
-                ->filter(function(Stopover $stopover) use ($departure, $startStation, &$encounteredStart): bool {
+                ->filter(function(Stopover $stopover) use ($startStation, &$encounteredStart): bool {
                     if (!$encounteredStart) { // this assumes stopovers being ordered correctly
-                        $encounteredStart = $stopover->departure_planned == $departure && $stopover->station->is($startStation);
+                        $encounteredStart = $stopover->station->is($startStation);
                         return false;
                     }
                     return true;
--- a/resources/vue/components/CheckinInterface.vue
+++ b/resources/vue/components/CheckinInterface.vue
@@ -38,7 +38,7 @@ export default {
                 business: this.business,
                 ibnr: true,
                 tripId: this.selectedTrain.tripId,
-                lineName: this.selectedTrain.line.name || this.selectedTrain.line.fahrtNr,
+                lineName: this.selectedTrain.line.name ?? this.selectedTrain.line.fahrtNr,
                 start: this.selectedTrain.stop.id,
                 destination: this.selectedDestination.evaIdentifier,
                 departure: DateTime.fromISO(this.selectedTrain.plannedWhen).setZone("UTC").toISO(),
--- a/resources/vue/components/CheckinInterface.vue
+++ b/resources/vue/components/CheckinInterface.vue
@@ -38,7 +38,7 @@ export default {
                 business: this.business,
                 ibnr: true,
                 tripId: this.selectedTrain.tripId,
-                lineName: this.selectedTrain.line.name,
+                lineName: this.selectedTrain.line.name || this.selectedTrain.line.fahrtNr,
                 start: this.selectedTrain.stop.id,
                 destination: this.selectedDestination.evaIdentifier,
                 departure: DateTime.fromISO(this.selectedTrain.plannedWhen).setZone("UTC").toISO(),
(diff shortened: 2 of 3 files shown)


app/Http/Controllers/FrontendTransportController.php
```diff
@@ -137,9 +138,9 @@ public function TrainTrip(Request $request): Renderable|RedirectResponse {
 
             $encounteredStart = false;
             $stopovers = $trip->stopovers
-                ->filter(function(Stopover $stopover) use ($departure, $startStation, &$encounteredStart): bool {
+                ->filter(function(Stopover $stopover) use ($startStation, &$encounteredStart): bool {
                     if (!$encounteredStart) { // this assumes stopovers being ordered correctly
-                        $encounteredStart = $stopover->departure_planned == $departure && $stopover->station->is($startStation);
```
MrKrisKrisu, line 142: Removing timestamps from station checks is not a good idea, while this will break ring lines (not checked, but in my experience this is the case every time...)

fabiancdng: I suppose you mean a scenario like this:

Screenshot 2024-03-10 at 19 15 12

That example works without the timestamp check based on the planned departure of the stopover. The filter only removes stopovers before the first occurrence of the station. After that it stops filtering out stations. Duplicate occurrences of the same station should, therefore, be fine. I might be thinking of the wrong scenario here though 😅

HerrLevin: Hmmm... I've checked it with the S41/S42 from Berlin Gesundbrunnen and it seems to work just fine. I guess we'll roll with it and revert it in a worst case scenario? :D

MrKrisKrisu: Sure. :D

MrKrisKrisu: @DC7IA do you have some input here before we are merging this as you worked on this snipped before?

MrKrisKrisu: @d3d9 do you have some input here before we are merging this as you worked on this snipped before?

HerrLevin:
@fabiancdng I'm fine with the changes. Close or resolve the comments with what works for you, give me a ping and we'll merge it.

HerrLevin:
@fabiancdng Wait no, one last thing: You might want to change all `||` to `??` because there might be some weird situation where a linename is `0` or sth. We're talking about public public transport - You have to expect the unexpected.

app/Http/Controllers/FrontendTransportController.php
fabiancdng: Reverting if something breaks is fine with me (of course). I'll mark this as resolved but feel free to gather more input if you like 🙂

fabiancdng:
@HerrLevin Good catch! 😄 I fixed the operator in [58e72ec](https://github.com/Traewelling/traewelling/pull/2407/commits/58e72ec11b9b0b3ddc646db50779547f24bdce7b).
Feel free to merge if there's nothing else.

HerrLevin (approved):

app/Http/Controllers/FrontendTransportController.php
k-a-vt: > @d3d9 do you have some input here before we are merging this as you worked on this snipped before?

Thank you for notifying me. I'm afraid I wasn't able to respond earlier.
I think the departure time check was there for a good reason. On my first attempt testing out the new code, I could notice how, also using the Gesundbrunnen example, with a train that is passing it the second time, I get shown the stopovers before the second Gesundbrunnen one. Checking in fails ("Departure time must be before arrival time"), but we should prevent the wrong stopovers being displayed.  
Instead of removing the departure time check, it should be kept and fixed with respect to timezones.

edit: oops, used the wrong account 😅

d3d9:
Here's a reminder that the removal of the check for departure time equality caused new issues on trips that contain the the same stopover multiple times, as I mentioned in https://github.com/Traewelling/traewelling/pull/2407#discussion_r1527346455 using my wrong account. I won't repeat the description, furthermore I don't have the capacity to create a new issue or fix it myself. I just wanted to put this comment here as a reminder since I feel it might have been overseen, especially as github hides it behind a resolved review.

PR #2408, opened 2024-03-10
:bug: fix StatusTagPolicy for unauthenticated users

### LINKED ISSUE #2335: Tags not publicly displayed, nor is visibility setting being applied
### Describe the bug

image

When you use these new tags, 
* they are not actually seen by not-logged in users – despite being set to public (by default)
* the visibility settings changes are not being applied

### Steps to reproduce

See video:
[Bildschirmaufzeichnung vom 2024-02-08, 17-55-43.webm](https://github.com/Traewelling/traewelling/assets/11966684/b768b265-fa5a-4479-9859-2e9465653939)

TL;DR/:In short:
1. Create a new tag and select a different visibility.
2. Bug one: It is not displayed as your category.
3. Try to change the category again manually.
   **Note:** Here, I noticed, no network request is even sent to the server.
   _Note:_ If it works (which one time it did, when I tried reproducing), just reload the page. As no information is sent to the server, it has the old visibility again.
4. Additionally have some public status and try to load that as an unauthenticated user. (Just switching tabs in the video, which you don't see and loading the URL.)
5. Bug two: No tags are displayed.

## More

Example https://traewelling.de/status/2004910

I have experimental features disabled, but had enabled it in the past.

### Browser console logs

```bash
N/A
```

### Browser

Firefox 122

MrKrisKrisu:
fixes #2335

### DIFF
--- a/app/Http/Controllers/Backend/Transport/StatusTagController.php
+++ b/app/Http/Controllers/Backend/Transport/StatusTagController.php
@@ -6,11 +6,12 @@ use App\Http\Controllers\Controller;
 use App\Models\Status;
 use App\Models\StatusTag;
 use App\Models\User;
+use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Gate;
 
 abstract class StatusTagController extends Controller
 {
-    public static function getVisibleTagsForUser(Status $status, User $user = null) {
+    public static function getVisibleTagsForUser(Status $status, User $user = null): Collection {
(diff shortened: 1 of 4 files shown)


jeyemwey (approved):

jeyemwey:
Would you be able to add unit tests for this Gate maybe? 

For the Status View Gate, I know that the unit tests have already saved our butts, so i like having them as a form of documentation.

jeyemwey (commented):
Besten Dank, sieht gut aus :3

Issue #2411, opened 2024-03-10
[Progress Tracker] Support for Manual Trips and more stations


MrKrisKrisu:
This pinned issue will document the progress of the project for Träwelling supporting manual trips and other stations
then known by "Deutsche Bahn HAFAS".
The goal is to change Träwelling this way, we are able to support multiple data sources.
To achieve this we want to use data from [Wikidata](https://www.wikimedia.de/projects/wikidata/) and the IFOPT ([Wikipedia](https://en.wikipedia.org/wiki/Identification_of_Fixed_Objects_in_Public_Transport), [OpenStreetMap Wiki](https://wiki.openstreetmap.org/wiki/DE:Key:ref:IFOPT)) as a unique identifier for stations.

> [!IMPORTANT]  
> Dear API Consumers, in the next few weeks and months we will change some endpoints and the data structure of Träwelling.
> Please be aware of this and check the [changelog](https://github.com/Traewelling/traewelling/blob/develop/API_CHANGELOG.md) regularly as there will be breaking changes.
> We try to keep the impact as low as possible and add backwards compatibility where it is possible.

> [!TIP]
> If you enjoy maintaining data, Wikidata is a suitable platform to save structured data so that a large number of projects can benefit from it under a free license.
> Many train stations, bus stops and more do not exist there or have not stored any associated IDs (IBNR, IFOPT/DHID, ...). Wikidata lives from people who volunteer data.
> If you are interested in helping to improve the data quality of public transport stations, you can have a look at train stations in your area and add missing information.
> Have a look at [Karlsruhe Main Station (Q688541)](https://www.wikidata.org/wiki/Q688541) as an example.

### Manual trip creation

- [x] Let `closed beta` users create trips manually with existing station data
    - [x] Check if there are any error on frontend, backend and API side
    - [x] Move trip creation from `closed beta` to `experimental features` (#2427 )

### Station support

- [x] Add Columns for storing the Wikidata ID in the `stations` table (#2389)
- [x] Link local station entity with Wikidata entity and fetch raw data for testing (#2389)
- [x] Add Columns for storing the IFOPT ID in the `stations` table (every segment of the IFOPT ID should be stored in a
  separate column!) (#2413)
- [x] Fetch IFOPT from Wikidata if available
- [x] Let users create stations from Wikidata Entities

### Detach from using the IBNR as a unique identifier

- [x] Change all API endpoints which are currently using the IBNR or the Name of a station as query to use the
  Träwelling ID
[...]

PR #2413, opened 2024-03-10
:card_file_box: add database layout for IFOPT

MrKrisKrisu:
see #2411


jeyemwey:
Why is it necessary to have 5 fields in the database for this? IMO it would be sufficient to keep it in one column with `:` in between.

HerrLevin:
@jeyemwey It's not necessary per say, but you can leverage the relational characteristics of the individual segments. That way, we can group stations based on segment a, b and c to use another API or stuff like that. Doing that with LIKE will be way less efficient.

app/Models/Station.php
```diff
@@ -29,22 +29,45 @@ class Station extends Model
     use HasFactory, LogsActivity;
 
     protected $table    = 'train_stations';
-    protected $fillable = ['ibnr', 'wikidata_id', 'rilIdentifier', 'name', 'latitude', 'longitude', 'time_offset', 'shift_time'];
+    protected $fillable = [
+        'ibnr', 'wikidata_id', 'rilIdentifier',
+        'ifopt_a', 'ifopt_b', 'ifopt_c', 'ifopt_d', 'ifopt_e',
+        'name', 'latitude', 'longitude', 'time_offset', 'shift_time'
+    ];
     protected $hidden   = ['created_at', 'updated_at', 'time_offset', 'shift_time'];
```
HerrLevin, line 37: Maybe add the individual ifopt segments to hidden?

database/migrations/2024_03_10_211526_add_ifopt_to_stations.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void {
+        Schema::table('train_stations', static function(Blueprint $table) {
+            $table->unsignedInteger('ifopt_e')->nullable()->comment('Platform')->after('wikidata_id');
+            $table->unsignedInteger('ifopt_d')->nullable()->comment('Area')->after('wikidata_id');
+            $table->unsignedInteger('ifopt_c')->nullable()->comment('Station')->after('wikidata_id');
+            $table->unsignedInteger('ifopt_b')->nullable()->comment('Municipal')->after('wikidata_id');
```
HerrLevin:
```suggestion
            $table->unsignedInteger('ifopt_b')->nullable()->comment('Administrative Area')->after('wikidata_id');
```

database/migrations/2024_03_10_211526_add_ifopt_to_stations.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void {
+        Schema::table('train_stations', static function(Blueprint $table) {
+            $table->unsignedInteger('ifopt_e')->nullable()->comment('Platform')->after('wikidata_id');
+            $table->unsignedInteger('ifopt_d')->nullable()->comment('Area')->after('wikidata_id');
+            $table->unsignedInteger('ifopt_c')->nullable()->comment('Station')->after('wikidata_id');
```
HerrLevin:
```suggestion
            $table->unsignedInteger('ifopt_c')->nullable()->comment('Mode / Stop Place')->after('wikidata_id');
```

database/migrations/2024_03_10_211526_add_ifopt_to_stations.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void {
+        Schema::table('train_stations', static function(Blueprint $table) {
+            $table->unsignedInteger('ifopt_e')->nullable()->comment('Platform')->after('wikidata_id');
+            $table->unsignedInteger('ifopt_d')->nullable()->comment('Area')->after('wikidata_id');
```
HerrLevin:
```suggestion
            $table->unsignedInteger('ifopt_d')->nullable()->comment('Stop place component')->after('wikidata_id');
```

HerrLevin:
```suggestion
            $table->unsignedInteger('ifopt_d')->nullable()->comment('Stop Place / Stop Place Component')->after('wikidata_id');
```

database/migrations/2024_03_10_211526_add_ifopt_to_stations.php
```diff
@@ -0,0 +1,31 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void {
+        Schema::table('train_stations', static function(Blueprint $table) {
+            $table->unsignedInteger('ifopt_e')->nullable()->comment('Platform')->after('wikidata_id');
```
HerrLevin:
```suggestion
            $table->unsignedInteger('ifopt_e')->nullable()->comment('Stop Place Component / unused')->after('wikidata_id');
```

database/migrations/2024_03_10_211526_add_ifopt_to_stations.php
HerrLevin: See http://www.normes-donnees-tc.org/wp-content/uploads/2014/05/CEN_TS_28701.pdf Page ~111

image

HerrLevin (changes requested):

app/Models/Station.php
MrKrisKrisu: Why? We don't expose Stations to API consumers, since we are using Resources. So I don't see a reason to use the $hidden property.

HerrLevin: Yeah you're right.

jeyemwey:
> you can leverage the relational characteristics of the individual segments. That way, we can group stations based on segment a, b and c to use another API or stuff like that. Doing that with LIKE will be way less efficient.

Jetzt wo ich drüber nachdenke, kann ich die Chancen sehen. Allein die Chance, Country-Informationen darzustellen, und zwar abseits von Ziffern in einer IBNR, ist ja schon nice. :3

MrKrisKrisu:
> It's not necessary per say

In the context of **database normalization**, I would argue that this is the most sensible approach. Since each attribute provides its own context, it is also easier to select - and I also see improved performance if we later want to search by country or municipal unit, for example.

HerrLevin (approved):

PR #2445, opened 2024-03-24
:technologist: move permission from migrations to seeder

MrKrisKrisu:


database/migrations/2024_02_01_000000_add_permissions_for_manual_trips_beta.php
```diff
@@ -1,25 +1,11 @@
...
-    }
-
-    public function down(): void {
-        $permission = Permission::findByName('create-manual-trip');
-
-        Role::findByName('admin')->revokePermissionTo($permission);
-        Role::findByName('closed-beta')->revokePermissionTo($permission);
-
-        $permission->delete();
+        //empty now - not deleted so a revert doesn't break
```
HerrLevin: image

You can apparently delete migrations without any problems. Have you tested what happens in case of a rollback with these specific migrations?

HerrLevin (approved):
Please test if a rollback would still work without the migrations. If so, please delete them to reduce useless files in the code. Otherwise it looks good to me.

PR #2501, opened 2024-04-16
:bug: Server error on Prometheus endpoint if cache is reset

HerrLevin:
Restructure cache access for static monitoring keys + add recreation of monitoring keys if absent


app/Enum/CacheKey.php
```diff
@@ -8,6 +8,14 @@
 
 enum CacheKey: string
 {
+    // static keys with fixed values
+    public const string STATUS_CREATED = 'monitoring-counter-StatusCreated';
+    public const string STATUS_DELETED = 'monitoring-counter-StatusDeleted';
+    public const string USER_CREATED = 'monitoring-counter-UserCreated';
```
MrKrisKrisu: please format <3

app/Enum/CacheKey.php
```diff
@@ -8,6 +8,14 @@
...
+    public const string USER_DELETED   = 'monitoring-counter-UserDeleted';
+    public const string WEBHOOK_ABSENT = 'monitoring-counter-WebhookAbsent';
```
MrKrisKrisu: Why are you using `public const string` here instead of `case`?

MrKrisKrisu (changes requested):

app/Enum/CacheKey.php
HerrLevin: Because I've reduced the need for an enum. That way we can directly use the key out of this class rather than generating a key with multiple method calls.

I don't think an enum was fitting here because we don't need an interface to check that one key is inside a specific range.

HerrLevin: Ah. Ooops. Now I get what you're saying. I kinda thought the original CacheKey.php was a Class rather than an enum. Let me fix that. 😅

PR #2505, opened 2024-04-19
✨ Reduce Brouter Polylines

HerrLevin:
Catch if another brouter Polyline is already set.

### DIFF
--- a/database/migrations/2024_04_19_131906_add_index_to_poly_line_parent_id.php
+++ b/database/migrations/2024_04_19_131906_add_index_to_poly_line_parent_id.php
@@ -6,16 +6,14 @@ use Illuminate\Support\Facades\Schema;
 
 return new class extends Migration
 {
-    public function up(): void
-    {
-        Schema::table('poly_lines', function (Blueprint $table) {
+    public function up(): void {
+        Schema::table('poly_lines', static function (Blueprint $table) {
             $table->index('parent_id');
         });
     }
 
-    public function down(): void
-    {
-        Schema::table('poly_lines', function (Blueprint $table) {
+    public function down(): void {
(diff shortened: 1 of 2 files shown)


database/migrations/2024_04_19_131906_add_index_to_poly_line_parent_id.php
```diff
@@ -0,0 +1,22 @@
...
+            $table->index('parent_id');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('poly_lines', function (Blueprint $table) {
+            $table->dropIndex('parent_id');
+        });
+    }
```
MrKrisKrisu:
```suggestion
    public function up(): void {
        Schema::table('poly_lines', static function (Blueprint $table) {
            $table->index('parent_id');
        });
    }

    public function down(): void {
        Schema::table('poly_lines', static function (Blueprint $table) {
            $table->dropIndex('parent_id');
        });
    }
```

database/migrations/2024_04_19_131906_add_index_to_poly_line_parent_id.php
```diff
@@ -0,0 +1,22 @@
...
+    {
+        Schema::table('poly_lines', function (Blueprint $table) {
+            $table->index('parent_id');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('poly_lines', function (Blueprint $table) {
+            $table->dropIndex('parent_id');
```
MrKrisKrisu: Did you checked, that reverting is working? 
As I know giving a string here would look for a index with the given name. I'd use `['parent_id']` as argument here.

MrKrisKrisu (approved):
If my question is answered/checked this count as an approve. :)

PR #2506, opened 2024-04-19
:bug: Reduce useless polyline nesting

HerrLevin:
Catch if another brouter Polyline is already set as parent.

### DIFF
--- a/app/Http/Controllers/Backend/BrouterController.php
+++ b/app/Http/Controllers/Backend/BrouterController.php
@@ -212,7 +212,7 @@ abstract class BrouterController extends Controller
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Backend/BrouterController.php
```diff
@@ -211,4 +211,19 @@ private static function checkIfPolylineHasMissingParts(Trip $trip): bool {
         }
         return false;
     }
+
+    private static function getOldPolyline(Trip $trip) {
```
MrKrisKrisu: return type?

MrKrisKrisu (approved):

PR #2530, opened 2024-04-26
:sparkles: Add interface for setting a user's permissions

HerrLevin:
image


app/Http/Controllers/Frontend/Admin/UserController.php
```diff
@@ -63,4 +64,20 @@ public function updateMail(Request $request): RedirectResponse {
         }
         return redirect()->route('admin.users.user', ['id' => $validated['id']]);
     }
+
+    public function updateRoles(Request $request): RedirectResponse {
```
MrKrisKrisu, line 77: I know this route is secured in the route file, but maybe add an additional security check here (double check for permission)?

MrKrisKrisu: (or maybe disable the NEW assigment to the admin role - only if user have it already. Setting new admins shouldn't be that easy.)

jeyemwey: > but maybe add an additional security check here (double check for permission)?

This is also a great test case, so there is a definitive answer if a code change breaks this permissions structure.

PR #2531, opened 2024-04-26
✨ Add weekly command for cleaning up mastodon servers

HerrLevin:
(this __might__ spam some users' notifications. Could also set it to monthly)

### DIFF
--- a/app/Console/Commands/CleanUpMastodonServers.php
+++ b/app/Console/Commands/CleanUpMastodonServers.php
@@ -1,0 +1,146 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\MastodonServer;
+use App\Models\User;
+use App\Notifications\InvalidMastodonServer;
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\GuzzleException;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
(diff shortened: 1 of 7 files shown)


app/Console/Commands/CleanUpMastodonServers.php
```diff
@@ -0,0 +1,146 @@
...
+class CleanUpMastodonServers extends Command
+{
+    protected $signature = 'trwl:clean-up-mastodon-servers';
+
+    protected $description = 'Clean up Mastodon servers...';
+    private const string DIVIDER       = '====================';
+    private const string SMALL_DIVIDER = '-------------------';
+
+
+    public function handle(): void {
```
MrKrisKrisu, line 23: Idea: In order to be able to track the changes made here, I would log the actions so that we can see this in the admin panel.

See here for easy examples: https://spatie.be/docs/laravel-activitylog/v4/basic-usage/logging-activity

MrKrisKrisu (approved):
my comment is just nice to have. :) looks good

app/Console/Commands/CleanUpMastodonServers.php
HerrLevin: I might add this down the road. Or feel free to add it yourself. :)

PR #2571, opened 2024-05-20
:sparkles: Add back-button functionality to Experimental Check-In (+ working direct links)

HerrLevin:


app/Http/Controllers/Frontend/VueFrontendController.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace App\Http\Controllers\Frontend;
+
+use Illuminate\Contracts\View\Factory;
+use Illuminate\Contracts\View\View;
+use Illuminate\Foundation\Application;
+
+class VueFrontendController
+{
+    public function stationboard(): Factory|Application|View|\Illuminate\Contracts\Foundation\Application {
```
MrKrisKrisu: return types are looking wrong. should be just `use Illuminate\View\View;`.

MrKrisKrisu (approved):

PR #2576, opened 2024-05-21
:loud_sound: add logging and delete after 14 days

MrKrisKrisu:
image


MrKrisKrisu:
will continue this later, we currently have other priorities.

PR #2617, opened 2024-05-29
:bug: Delete old notifications based on created_at

HerrLevin:


MrKrisKrisu (approved):

MrKrisKrisu:
maybe also reduce the time from 6 to 2 months?

MrKrisKrisu:
Or less. Maybe 14 days can ca enough because all notifications are "Adhoc notifications".

HerrLevin:
> Or less. Maybe 14 days

Also for the already read notifications?

MrKrisKrisu:
> Also for the already read notifications?

Yes, I think we can delete all notifications 14 days after created.

PR #2621, opened 2024-05-30
:recycle: rename `begin` and `end` to better distinguish

MrKrisKrisu:


app/Notifications/EventSuggestionProcessed.php
```diff
@@ -32,9 +32,9 @@ public function via(): array {
 
     public function toArray(): array {
         return [
-            'accepted'      => $this->event !== null,
-            'event'         => $this->event?->only(['id', 'slug', 'name', 'begin', 'end']),
-            'suggestedName' => $this->eventSuggestion->name,
+            'accepted'        => $this->event !== null,
+            'event'           => $this->event?->only(['id', 'slug', 'name', 'checkin_start', 'checkin_end']),
```
marhei, line 36: Breaking API change without information 👀

marhei: I mean when you already changed the keys on this API endpoint you should do it everywhere.

MrKrisKrisu: Oh, well. This wasn't documented (and not tested), so I didn't saw that this is also provided to API.

marhei: Yes, the returned notification API data is not documented at all. It only says: Just try it!

PR #2625, opened 2024-05-30
:recycle: rename `trainDistance` and `trainDuration` to `totalDistance` and `totalDuration`

MrKrisKrisu:
also:
- minify api docs
- add `roles` to UserAuthResource (/api/v1/auth/user)


MrKrisKrisu:
> [!IMPORTANT]  
> Renamed `trainDuration` and `trainDistance` attributes to `totalDuration` and `totalDistance` in all User object. (We have more than just trains.)
> 
> The old attributes are still available and will be removed after 2024-08.

(@marhei, @vainamov, @bendix-dev, @jheubuch, @timoschwarzer as of #2619)

PR #2635, opened 2024-06-01
:technologist: Replace user details by `LightUserResource` in `LeaderboardUserResource`

marhei:
To provide the same basic informations across all user resources


marhei:
Maybe some also needs to recreate the API docs, I don't know how this works…

storage/api-docs/api-docs.json
```diff
@@ -18,7 +18,7 @@
             "description": "Production Server"
         },
         {
-            "url": "http://localhost:8000/api/v1",
+            "url": "http://localhost:8000/api/v1/api/v1",
```
marhei: I don't get it why this happened?

app/Http/Resources/LeaderboardUserResource.php
```diff
@@ -19,7 +21,9 @@ class LeaderboardUserResource extends JsonResource
 {
     public function toArray($request): array {
         return [
-            'username'       => $this->user->username,
+            'id'             => (int) $this->user->id,
+            'displayName'    => (string) $this->user->name,
+            'username'       => (string) $this->user->username,
             'profilePicture' => ProfilePictureController::getUrl($this->user),
```
MrKrisKrisu: Maybe we can include the LightUserResource here. This reduced the amount of different resources with similar attributes.

What do you think?

storage/api-docs/api-docs.json
```diff
@@ -5944,9 +5952,9 @@
                 "scheme": "https",
                 "flows": {
                     "authorizationCode": {
-                        "authorizationUrl": "http://localhost:8000/oauth/authorize",
-                        "tokenUrl": "http://localhost:8000/oauth/token",
-                        "refreshUrl": "http://localhost:8000/auth/refresh",
+                        "authorizationUrl": "http://localhost:8000/api/v1/oauth/authorize",
+                        "tokenUrl": "http://localhost:8000/api/v1/oauth/token",
+                        "refreshUrl": "http://localhost:8000/api/v1/auth/refresh",
```
MrKrisKrisu: ah, well. now we have this...

MrKrisKrisu: @HerrLevin Can you help me here before I break something, you've done?

storage/api-docs/api-docs.json
MrKrisKrisu: This is resolved... but I've created another issue (see below). :c

//cc @HerrLevin

app/Http/Resources/LeaderboardUserResource.php
marhei: I would prefer my proposed implementation because it fits better into my existing implementation of the user model.

storage/api-docs/api-docs.json
marhei: I think that's right?

marhei:
@MrKrisKrisu As you wanted: `LeaderboardUserResource` returns now a `LightUserResource`

marhei:
I don't know why the test doesn't want to run. 😤

MrKrisKrisu (approved):
Thanks for the change. But maybe we should get a third opinion before we just blindly change something just because I like it more?

(@HerrLevin?)

marhei:
@HerrLevin May I be merged? 🥺👉👈

config/l5-swagger.php
```diff
@@ -288,7 +288,7 @@
          */
         'constants'             => [
             'L5_SWAGGER_CONST_HOST' => env('L5_SWAGGER_CONST_HOST', 'http://localhost:8000'),
-            'L5_SWAGGER_CONST_ENDPOINT' => env('L5_SWAGGER_CONST_HOST', 'http://localhost:8000') . '/api/v1',
```
HerrLevin: This needs to be changed back because you won't be able to test the api locally via swagger.

Screenshot 2024-07-02 at 12 39 26

storage/api-docs/api-docs.json
```diff
@@ -18,7 +18,7 @@
             "description": "Production Server"
         },
         {
-            "url": "https://traewelling.de/api/v1/api/v1",
+            "url": "http://localhost:8000",
```
HerrLevin: See last comment

HerrLevin (changes requested):

HerrLevin:
@marhei I know where you're coming from. From a client pov I would prefer the original user pov. But I know our code and our database... This would most likely impact our performance (which we don't have :D) severely 🥴

I'm fine with the PR as it is currently, if you revert the two code occurences I've mentioned in my review. /cc @MrKrisKrisu

marhei:
I removed all unwanted changes, thanks for the review!

HerrLevin (approved):

PR #2639, opened 2024-06-01
♻️ Adapt operator(s) endpoint to adhere to zalando convention

HerrLevin:
The endpoint `/operator` will be changed to `/operators`

> [!NOTE]
> The renamed endpoint was only set up yesterday, so it doesn't really affect anyone - unless you're really really early.


HerrLevin:
@Traewelling/api-consumer

PR #2658, opened 2024-06-04
:children_crossing: update social post length to 500 chars

karlbeecken:
As Twitter social posting is not working anymore, the maximum post length can be upped to 500 chars, the default maximum length for a Mastodon post.
Also updated the `...` to a true unicode `…`, saving some more chars.

### DIFF
--- a/app/Http/Controllers/Backend/Helper/StatusHelper.php
+++ b/app/Http/Controllers/Backend/Helper/StatusHelper.php
@@ -66,9 +66,9 @@ class StatusHelper
             $body           = $this->mastodon ? $this->getMastodonBody() : $this->status->body;
             $appendix       = $this->generateAppendix();
             $appendixLength = strlen($appendix) + 30;
-            $postText       = substr($body, 0, 280 - $appendixLength);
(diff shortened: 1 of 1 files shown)


karlbeecken:
One could also up the maximum status text length to 500 (currently also limited to 280), but this may not be wanted to keep things brief.

app/Http/Controllers/Backend/Helper/StatusHelper.php
```diff
@@ -66,9 +66,9 @@ private function generateSocialText(): string {
             $body           = $this->mastodon ? $this->getMastodonBody() : $this->status->body;
             $appendix       = $this->generateAppendix();
             $appendixLength = strlen($appendix) + 30;
-            $postText       = substr($body, 0, 280 - $appendixLength);
+            $postText       = substr($body, 0, 500 - $appendixLength);
```
MrKrisKrisu, line 69: This change would not change anything, as the status body cannot be longer than 280 characters.

MrKrisKrisu (approved):

HerrLevin (approved):

PR #2777, opened 2024-07-16
:globe_with_meridians: Fix translations. The map provider is CarTo, not carGo

pietervdvn:
Seems like you rail fanatics write more about cargo trains then maps ("carto") ;)

I originally wanted to change all mentions of cargo, but I guess that'll cause issues with the database, so I sticked with only changing the user-facing translations.


MrKrisKrisu (approved):
image

MrKrisKrisu:
Thank you! I'll keep my eyes open in the code bit by bit and also change the remaining parts.

(It's a bit funny though)

Issue #2551, opened 2024-05-08
API: Inconsistent use of camel and snake case in request body


marhei:
To simplify the use of automatic key encoding strategies, a standardised use of either camel or snake case for keys in the request body would make sense. Currently, for example, camel case is used in the `/trains/checkin` endpoint, but snake case is used in the new `/report` endpoint.

MrKrisKrisu:
In long term I would have a look at the API Guidelines from Zalando which recommend [kebab-case](https://opensource.zalando.com/restful-api-guidelines/#129). (idea from @HerrLevin)

https://opensource.zalando.com/restful-api-guidelines/

MrKrisKrisu:
Oh, sorry. You mean the keys. I don't know if there is a recommendation, but I personally prefer camelCase here. We can discuss that here.

marhei:
As all other endpoints use keys in camel case and the report endpoint is not yet widely used, I would recommend changing the key from this endpoint to camel case as well.

PR #2781, opened 2024-07-17
🐛  Fix /reports endpoint

HerrLevin:


HerrLevin:
@Traewelling/api-consumer This PR contains a breaking change of the `/report` endpoint. The change will be breaking after 2024-08-17

jheubuch:
Thanks :)

marhei:
This will also fix #2551, thanks for the change!

HerrLevin:
> This will also fix #2551, thanks for the change!

Sadly not completely. We still have `likes_enabled` in the user model.

PR #2792, opened 2024-07-24
🚸 Move regionalExp to the group of long-distance-trains

HerrLevin:
We've been doing it wrong for way too many years and I'm sick of it.

regionalExp is NOT the category for RE, IRE, etc (looking at you DB-Rest 👀)

It's simply the category for [long-distance trains / D-Trains](https://en.wikipedia.org/wiki/Schnellzug) like Flix-Train, Westbahn, etc.

If you look at stations that ONLY run regional transport, you'll see a `"regionalExp": false`
Screenshot 2024-07-24 at 23 11 21

WB:
Screenshot 2024-07-24 at 23 09 17
FLX:
Screenshot 2024-07-24 at 23 09 55

### DIFF
--- a/app/Enum/HafasTravelType.php
+++ b/app/Enum/HafasTravelType.php
@@ -29,16 +29,17 @@ enum HafasTravelType: string
 
     public function getEmoji(): string {
         return match ($this->value) {
-            'nationalExpress', 'national' => '🚄',
-            'regionalExp', 'regional'     => '🚆',
-            'suburban'                    => '🚋',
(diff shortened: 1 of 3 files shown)


d3d9:
This would be great, especially considering the current DB Fernverkehr replacement buses which also are of this type.

Aren't there further changes required e.g. for the filter buttons or statistics view? See https://github.com/Traewelling/traewelling/discussions/2364#discussioncomment-8506488

MrKrisKrisu (approved):

PR #2885, opened 2024-08-30
:bug: Intercitys are not being shown with the long distance filter

HerrLevin:
TravelType::EXPRESS now covers HTT::REGIONAL_EXP


HerrLevin:
I think @marhei and Lokbuch's release notes will like this 👀

marhei:
As we filter locally, this will not change the behaviour of Lokbuch. But thanks anyway... :D

PR #2932, opened 2024-09-29
:passport_control: add role `deactivate-account-usage`

MrKrisKrisu:
With the new role `deactivate-account-usage` we can now:

- disallow the creation of manual trips
- disallow the creation of checkins/statuses
- disallow the updating of non-private statuses (user can still set existing statuses to private)
- disallow social interaction (following, liking)

This role is a step we don't want to take, but we have to. We actually want to intervene as little as possible in users' accounts, but the past has shown that this is unfortunately sometimes necessary. We try to use this intervention as rarely as possible and use less stringent interventions first.

Deactivating the basic account functions represents the greatest possible restriction for us before deleting user accounts.

### DIFF
--- a/app/Http/Controllers/API/v1/FollowController.php
+++ b/app/Http/Controllers/API/v1/FollowController.php
@@ -54,6 +54,7 @@ class FollowController extends Controller
      */
     public function createFollow(int $userId): JsonResponse {
         try {
+            $this->authorize('create', Follow::class);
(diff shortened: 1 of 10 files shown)


MrKrisKrisu:
When a user tries to check in, they receive an error message. The UI has not been customized here.

image


---------------------
If a user tries to update statuses that are not private, this is prevented and the user is asked to set the status to private.

image

PR #2948, opened 2024-10-10
feat(ui): add _Update Station Wikidata_ button

weiland:
This PR aims to help with updating existing station information. If there is new data in Wikidata, there is no way of updating this data in the (normal) user UI.

Solution/idea:
This PR adds another button with the ability to re-fetch and update the station information with new data from Wikidata.

To achieve this the GET parameter `allowUpdate` (`boolean`) was added to the `POST /api/v1/experimental/wikidata?allowUpdate=1` endpoint. This parameter prevents the early exit (because the `wikidata_id` already exists) and so `WikidataImportService::searchStation` is invoked, which updates the Station database with fresh data from Wikidata.

I could locally test this feature successfully.

There are some more questions:

- is this feature (request) useful? (if not, just stop reading and close the PR)
- is a GET parameter the correct choice here? (or would another endpoint suit better)
   - is the name `allowUpdate` alright?
- is the `allowUpdate` parameter even required or can it's behaviour be the default?
- what is the best way to prevent code duplication for the JS function `fetchWikidata` (and where to put it then)?

I'm looking forward for some input. This PR could be a starting point, it this feature is desired and I would be open to work it, if desired.

### DIFF
--- a/app/Http/Controllers/API/v1/ExperimentalController.php
+++ b/app/Http/Controllers/API/v1/ExperimentalController.php
@@ -23,8 +23,10 @@ class ExperimentalController extends Controller
             return response()->json(['error' => 'This station was already requested recently. Please try again later.'], 429);
         }
 
+        $allowUpdate = request()->boolean('allowUpdate');
+        // TODO: there will be an error when updating non-exsiting stations.
         $station = Station::findOrFail($stationId);
-        if ($station->wikidata_id) {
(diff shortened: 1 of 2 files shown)


MrKrisKrisu (approved):
🥰🥰🥰

MrKrisKrisu:
> * is a GET parameter the correct choice here? (or would another endpoint suit better)

As this is experimental and not documented:
We can test and if required change it in the future.

PR #2985, opened 2024-11-06
:sparkles: Add `followedBy` attribute to `UserResource`

jheubuch:
This PR adds a `followedBy` attribute to the `UserResource` which is set to true if the requested user is following the requesting user.


jheubuch:
I also added some UI:

image

image

MrKrisKrisu (approved):

PR #3000, opened 2024-11-20
✨ Extract Domain formatter + add more formats to accept

HerrLevin:


app/Http/Controllers/Backend/Social/MastodonController.php
```diff
@@ -54,7 +55,7 @@ public static function getUserFromSocialite(SocialiteUser $socialiteUser, Mastod
      * @throws InvalidMastodonException
      */
     public static function getMastodonServer(string $domain): ?MastodonServer {
-        $domain = self::formatDomain($domain);
+        $domain = (new MastodonDomainExtractionService())->formatDomain($domain);
```
MrKrisKrisu, line 58: maybe wait for us to be php 8.4 only and you can remove the additional `(..)`?

HerrLevin: I could also remove it once we're on 8.4 🙈

MrKrisKrisu: was just an idea, do what you want here :D

PR #3061, opened 2024-12-27
✨ add seehafen

jonasfranz:
Maybe we should find a better way to handle inherited types since seaport is instance of port. But for example Port of Barcelona is a seaport and only transitively a port.

### DIFF
--- a/app/Services/Wikidata/WikidataImportService.php
+++ b/app/Services/Wikidata/WikidataImportService.php
@@ -24,6 +24,7 @@ class WikidataImportService
         'Q953806', // Bushaltestelle
         'Q2175765', // Straßenbahnhaltestelle
         'Q44782', // Hafen
+        'Q15310171', // Seehafen
     ];
 
     public static function importStation(string $qId): Station {


MrKrisKrisu:
> Maybe we should find a better way to handle inherited types since seaport is instance of port. But for example Port of Barcelona is a seaport and only transitively a port.

It certainly makes sense in the long term. At the moment, however, I think it's still okay because it's a bit experimental.

MrKrisKrisu (approved):

jonasfranz:
@MrKrisKrisu I have a idea for fixing this problem. There is a option to use subclassOf instead of instanceOf. But this would require using SparQL instaead of just fetching the json.

this for example returns all instances of ports (including seaport etc.):
```sparql
#Cats
SELECT ?item ?itemLabel
WHERE
{
  ?item wdt:P31/wdt:P279* wd:Q44782. # Must be a cat
  SERVICE wikibase:label { bd:serviceParam wikibase:language "[AUTO_LANGUAGE],mul,en". } # Helps get the label in your language, if not, then default for all languages, then en language
}
```

MrKrisKrisu:
> @MrKrisKrisu I have a idea for fixing this problem. There is a option to use subclassOf instead of instanceOf. But this would require using SparQL instaead of just fetching the json.
> 
> this for example returns all instances of ports (including seaport etc.):
> 
> ```sparql
> #Cats
> SELECT ?item ?itemLabel
> WHERE
> {
>   ?item wdt:P31/wdt:P279* wd:Q44782. # Must be a cat
>   SERVICE wikibase:label { bd:serviceParam wikibase:language "[AUTO_LANGUAGE],mul,en". } # Helps get the label in your language, if not, then default for all languages, then en language
> }
> ```

Feel free! Would be nice. :)

PR #3069, opened 2024-12-29
:sparkles: add operator to TripCreationForm

MrKrisKrisu:


kasmar00:
Hi, small note: the `/api/v1/operators` endpoint has paging, and this loads only the first 100 operators:
image

Also `trip_creation.form.operator` has no translation 😉 
image

(I don't think this is worth submitting a bug as this form is work in progress, but correct me if I'm wrong 😄. Thanks for a great product 😉)

MrKrisKrisu:
Well - that happens when you do something very quick and dirty. Thanks for the hint!

SpielenmitLili:
Vielleicht kann man jetzt auch diese Info entfernen? 
Sie ist ja mit diesem Commit nicht mehr zutreffend

image

PR #3074, opened 2024-12-31
🔥 Remove IDE-specific configurations and prevent future commits

MrKrisKrisu:
This pull request removes IDE-specific configuration files (.idea, .vscode) from the repository to maintain a cleaner and more consistent codebase.

**Changes:**
🔥 Removed existing IDE configuration files from the repository.
🙈 Updated .gitignore to prevent committing IDE-specific files in the future.

**Why:**
IDE-specific files are personal to individual developers and should not be included in the shared repository. This ensures a more universal setup across different development environments and avoids unnecessary conflicts or clutter.

**Discussion:**
For cross-project settings, it's better to rely on standard tools like .editorconfig to define consistent coding styles and configurations. This approach ensures all contributors can follow the same standards, regardless of their development environment.

> [!IMPORTANT]
> Before pulling the changes here, make sure to back up your local IDE settings if needed. The removal of these files from the repository means they will no longer be available, and you might lose custom configurations that are important for your workflow. Save them locally to avoid any disruptions!


HerrLevin:
I'm fine with this, if you extend the .editorconfig to the rules you're deleting now. Otherwise the autoformat will go haywire. :D

MrKrisKrisu:
> I'm fine with this, if you extend the .editorconfig to the rules you're deleting now. Otherwise the autoformat will go haywire. :D

Will do - can I do this in another PR?

MrKrisKrisu:
I've dumped the IDE settings to `.editorconfig` in #3078

MrKrisKrisu:
Announcement: https://github.com/Traewelling/traewelling/discussions/3079

PR #3108, opened 2025-01-09
Fix validation bug in TripController

luelista:
Currently, the departure is always required, even if an arrival is given.

### DIFF
--- a/app/Http/Controllers/API/v1/TripController.php
+++ b/app/Http/Controllers/API/v1/TripController.php
@@ -53,7 +53,7 @@ class TripController extends Controller
                 'stopovers.*.stationId'     => ['required', 'exists:train_stations,id'],
                 'stopovers.*.arrival'       => ['required_without:stopovers.*.departure', 'required_with:stopovers.*.arrivalReal', 'date'],
                 'stopovers.*.arrivalReal'   => ['nullable', 'date'],
-                'stopovers.*.departure'     => ['required_without:stopovers.*.arrival,null', 'required_with:stopovers.*.departureReal', 'date'],
+                'stopovers.*.departure'     => ['required_without:stopovers.*.arrival', 'required_with:stopovers.*.departureReal', 'date'],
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/API/v1/TripController.php
```diff
@@ -53,7 +53,7 @@ public function createTrip(Request $request): TripResource|JsonResponse {
                 'stopovers.*.stationId'     => ['required', 'exists:train_stations,id'],
                 'stopovers.*.arrival'       => ['required_without:stopovers.*.departure', 'required_with:stopovers.*.arrivalReal', 'date'],
                 'stopovers.*.arrivalReal'   => ['nullable', 'date'],
-                'stopovers.*.departure'     => ['required_without:stopovers.*.arrival,null', 'required_with:stopovers.*.departureReal', 'date'],
+                'stopovers.*.departure'     => ['required_without:stopovers.*.arrival', 'required_with:stopovers.*.departureReal', 'date'],
```
MrKrisKrisu, line 56: I don't see any fixed value here. If an arrival is specified, departure may already be empty. But one of the two is necessary.

MrKrisKrisu (changes requested):

MrKrisKrisu:
I'll close this PR for now. Let me know if you'd like to continue, and I'll reopen it.

app/Http/Controllers/API/v1/TripController.php
luelista: @MrKrisKrisu 

The `,null` is incorrect, probably just a typo. 

Adding a `,null` to a [required_without](https://laravel.com/docs/11.x/validation#rule-required-without) rule says: require this field in case: either `stopovers.*.arrival` is *empty or not present* or `null` is *empty or not present*. However, `null` is not a valid field name, so, by definition, it is always *empty or not present*. So putting `required_without:...,some_non_existing_field` is the same as just setting it to `required`. Which is not what we want here.

PR #3112, opened 2025-01-09
:lipstick: Register form fixes

gwehrle:
I added some usability fixes for register form:

- different id for username. Before id `name` was used for name and username. This is important that the correct label is used for the input.
- add validation of password length to frontend
- make url filed required and to be an url for registration with mastodon


MrKrisKrisu (approved):
Thanks!

PR #3129, opened 2025-01-11
:fire: remove global dashboard

MrKrisKrisu:
The global dashboard has not been working correctly for several years (loading times are far too long, sometimes several minutes). I therefore propose to remove it. 

### In this PR:
- Remove Global Dashboard from Traewelling Frontend
- Remove API Documentation for `/api/v1/dashboard/global`
  - deprecate endpoint (will be removed, after a date we choose here in discussion)


jeyemwey:
Agree with the sentiment, still sad to see it go. This is potentially important for @Traewelling/api-consumer even if the API for the global dashboard is still works.

jheubuch:
Since it was (personally) never a use case for me to have this global dashboard, I'm fine with removal. 🫣

PR #3165, opened 2025-01-17
:bug: Fix: use correct deletion function for inactive users

ygerlach:
I have seen the code deleting inactive users in the `AccountDeletionController`. It misses the parts of deleting a user as seen in `UserController::deleteUserAccount` (ProfilePicture and DatabaseNotification).

Im not a laravel expert, but i believe this might be a bug keeping ProfilePictures and Notifications for deleted users. If this problem is real, we may need to create a migration that checks the Database for non deleted profile pictures and notification objects, as this might be a privacy concern.

### DIFF
--- a/app/Http/Controllers/Backend/User/AccountDeletionController.php
+++ b/app/Http/Controllers/Backend/User/AccountDeletionController.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers\Backend\User;
 
 use App\Helpers\CacheKey;
 use App\Http\Controllers\Controller;
+use App\Http\Controllers\Backend\UserController;
 use App\Mail\AccountDeletionNotificationTwoWeeksBefore;
 use App\Models\User;
 use Exception;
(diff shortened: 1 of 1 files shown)


MrKrisKrisu:
> but i believe this might be a bug keeping ProfilePictures and Notifications for deleted users. If this problem is real, we may need to create a migration that checks the Database for non deleted profile pictures and notification objects, as this might be a privacy concern.

Notifications are automatically deleted after a certain amount of time. As for profile pictures, we have a script that periodically deletes any that are not associated with an account. So this should not be an acute problem.

https://github.com/Traewelling/traewelling/blob/3b6d54758f2dd59af92f8091ceb0394e855bbd64/app/Console/Commands/CleanUpProfilePictures.php#L15-L25

ygerlach:
ok, so the data should be gone anyways. But i guess deleting it directly wouldn't hurt?

MrKrisKrisu:
> ok, so the data should be gone anyways. But i guess deleting it directly wouldn't hurt?

Nope, should be done like this. Can you please have a look at the tests, as currently the code is broken and I cannot merge it.

Also - did you test it?

ygerlach:
> Also - did you test it?

No. As i didn't manage to get a working dev setup yet.

I was just posting this MR as a discussion question, i didn't expect such a fast response.

MrKrisKrisu (approved):
> > Also - did you test it?
> 
> No.

Works on my machine - thanks. :)

PR #3194, opened 2025-01-28
🐛 Define the type cast of the number attribute in trip model

marhei:
Currently, the type of the number attribute is not defined in the trip model, which leads to undocumented API returns as integer. To follow the documentation which defines number as a string, I added a type cast to the model.

### DIFF
--- a/app/Models/Trip.php
+++ b/app/Models/Trip.php
@@ -48,6 +48,7 @@ class Trip extends Model
     protected $casts    = [
         'id'             => 'integer',
         'trip_id'        => 'string',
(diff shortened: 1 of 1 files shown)


marhei:
This PR should fix this bug which occurs occasionally in Lokbuch while loading the stopovers.

image

MrKrisKrisu (approved):
As `number` is also a varchar(255) in the database (lol), it should be fine.

marhei:
Yes, cached trips returned a string, freshly requested from the railway API an integer.

PR #3416, opened 2025-04-17
♻️ remove legacy autocomplete endpoint

### LINKED ISSUE #3413: Search for Wikidata IDs does not work for stops or destination of custom journeys
### Describe the problem

Search for Wikidata IDs does not work when searching stops for in the "create journey" interface (for intermediate stops or the trip's destination)

### What steps did you take before the issue occurred?

1. In the dashboard, search for "Q1766008"
2. Select "Sol" station
3. Create a journey
4. As destination, enter "Q2474438"

### What did you expect to happen?

Being offered "Tirso de Molina" as selection for the destination

### Browser console or error messages (optional)

```bash

```

### Any additional hints or information? (optional)

It seems when the stop has been searched before, it's possible to search the Wikidata stop by its name, but if you're on a route where many stops have never been searched before, it's really cumbersome to search for them in the dashboard before being able to search them in the create journey interface

MrKrisKrisu:
closes #3413


app/Http/Controllers/Backend/Transport/StationController.php
```diff
@@ -70,9 +78,26 @@ public static function getAlternativeDestinationsForCheckin(Checkin $checkin): C
...
+        }
+
+        $stations = $this->dataProvider->getStations($search);
+        if ($stations->count() < 10) {
+            $remaining  = 10 - $stations->count();
+            $dbStations = $this->stationRepository->getStationByName($search, 'de', true, $remaining);
+            $stations   = $stations->merge($dbStations);
+        }
+        return $stations;
+/*
```
MrKrisKrisu: ToDo: Check if this is required

app/DataProviders/Motis.php
```diff
@@ -59,17 +60,12 @@ public function getStationByRilIdentifier(string $rilIdentifier): ?Station {
         return null;
     }
 
+    /**
+     * @deprecated
+     */
     public function getStationsByFuzzyRilIdentifier(string $rilIdentifier): Collection {
-        $stations = Station::where('rilIdentifier', 'LIKE', "$rilIdentifier%")
```
HerrLevin, line 63: this should've been moved into a repository rather than the controller

resources/js/components/station-autocomplete.js
```diff
@@ -40,7 +40,7 @@ const popularStations = [
     function fetchStations() {
         if (input.value.length < 2) return;
 
-        fetch("/transport/train/autocomplete/" + encodeURI(input.value))
+        fetch("/admin/transport/train/autocomplete/" + encodeURI(input.value))
```
MrKrisKrisu: is this temporary?

HerrLevin: Was stupid here. Thought this endpoint is only used in the admin pages but it's still used in the event page.

PR #3461, opened 2025-04-29
🐛 Fix looking for default area

MartinNemi03:
Noticed that city names have recently disappeared from areas in autocomplete and it seems that default boolean has been moved


HerrLevin:
Thanks for pointing this out! This is an error in our API and not in the frontend. The changed format wasn't intended. Fixed it in #3462.

PR #3565, opened 2025-05-22
🎨 Return Station objects in getLatestArrivals

MartinNemi03:
Originally meant to have this add areas to station history (thus having area in autocomplete before typing a destination) and even though it seems that attribute would just need accessing in order to write itself into the output object, it seemed kinda iffy.. 

So at least it now returns Station objects as commented in TODO.

### DIFF
--- a/app/Http/Controllers/Backend/Transport/StationController.php
+++ b/app/Http/Controllers/Backend/Transport/StationController.php
@@ -32,6 +32,7 @@ class StationController extends Controller
      * @return Collection
      */
     public static function getLatestArrivals(User $user, int $maxCount = 5): Collection {
+        $groupAndSelect = [
             'train_stations.id', 'train_stations.ibnr', 'train_stations.name',
             'train_stations.latitude', 'train_stations.longitude', 'train_stations.rilIdentifier',
         ];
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Backend/Transport/StationController.php
```diff
@@ -32,19 +32,17 @@ public function __construct(?StationRepository $stationRepository = null) {
...
+        return DB::table('train_checkins')
+                ->join('train_stopovers', 'train_checkins.destination_stopover_id', '=', 'train_stopovers.id')
+                ->select('train_stopovers.train_station_id as station_id')
+                ->where('train_checkins.user_id', $user->id)
+                ->groupBy('station_id')
+                ->orderByDesc(DB::raw('MAX(train_checkins.arrival)'))
+                ->limit($maxCount)
+                ->pluck('station_id')
+                ->map(function (int $id) {
+                    return Station::where('id', $id)->first();
```
MrKrisKrisu: This would result in an additional `$maxCounts` SQL queries, which would not scale well on the production instance. :(

MrKrisKrisu (changes requested):

app/Http/Controllers/Backend/Transport/StationController.php
MartinNemi03: That's fair and thanks for letting me know, this isn't the best solution in regards to performance as I wasn't really looking into how many queries that would add. I'll turn this into a draft PR and try to approach it in a different way.

MartinNemi03: Went ahead and just made it so it maps the values into a new Station instance.

HerrLevin (approved):

Issue #3706, opened 2025-07-15
Mention on the stationboard page that data is filtered


jbruechert:
### Describe the problem

People are quite regularly confused about why their departure is not listed. It would be nice to show them this possible reason, so they don't have to ask on the Transitous side and get sent back here.

It should be possible to only display this if departures were filtered.

### What steps did you take before the issue occurred?

​

### Browser console or error messages (optional)

```bash

```

HerrLevin:
I'm not quite sure if I'm following this correctly.

Do you mean some kind of banner with something along the lines of "Departures at 13:37, showing only long distance trains"?

Or rather a notice that we're only showing some data based on the currently allowed Licenses (and a link to the license page)?

jbruechert:
I'm referring to the licenses, yes. I don't think a banner is needed for options set by the user.

HerrLevin:
I've implemented it, added a (hopefully sufficient) explanation in our help pages and pinged all api-subscribers so that they can include this in their apps. 

<img width="771" height="504" alt="Image" src="https://github.com/user-attachments/assets/26b1de53-e1b3-426f-a976-dd0ca1db0f18" />

<img width="759" height="286" alt="Image" src="https://github.com/user-attachments/assets/fdeb30cf-5aa8-498b-905c-04e8fd9e9bb6" />

jbruechert:
Thank you very much, this looks exactly like what I was missing :)

jbruechert:
Is it possible this doesn't work for stations with no departures left at all? (For example Beograd centar)

PR #3710, opened 2025-07-16
✨Output information about removed departures due to licensing issues

HerrLevin:
@Traewelling/api-consumer there's new additional information in our departures-endpoint.
This might be relevant to some users.

Screenshot 2025-07-16 at 19 24 01

--

Closes #3710


HerrLevin:
Closes #3706

PR #3764, opened 2025-08-02
⬆️ Update flake inputs and their dependencies

jeyemwey:
This is necessary because Vite expects Node 22, and flake's nixpkgs only served Node 20. Now we're at a Node v22.17.0.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,24 +2,32 @@
   "nodes": {
     "cachix": {
       "inputs": {
-        "devenv": "devenv_2",
-        "flake-compat": "flake-compat_2",
+        "devenv": [
+          "devenv"
+        ],
+        "flake-compat": [
+          "devenv"
+        ],
+        "git-hooks": [
+          "devenv",
+          "git-hooks"
+        ],
         "nixpkgs": [
           "devenv",
           "nixpkgs"
-        ],
-        "pre-commit-hooks": "pre-commit-hooks"
+        ]
       },
       "locked": {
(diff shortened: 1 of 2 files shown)


jeyemwey:
Keine Ahnung, wer hier reviewen möchte, weil @NyCodeGHG und ich die einzigen beiden waren, die überhaupt das Flake benutzen (gehe ich von aus zumindest)

NyCodeGHG:
This breaks when I try to `nix develop --impure` on a fresh setup, doesn't break without this PR (still kind of because of stuff in .env.example), (but the flake could also use a refactoring).
```
error:
       … while evaluating an expression to select 'drvPath' on it
         at «internal»:1:552:
       … while evaluating strict
         at «internal»:1:552:
       (stack trace truncated; use '--show-trace' to show the full trace)

       error: attribute 'DB_DATABASE' missing
       at /nix/store/ri1pq88aymnbrbpsk0dmkmvwspxzm5qq-source/flake.nix:52:35:
           51|               {
           52|                 name = config.env.DB_DATABASE;
             |                                   ^
           53|               }
```

MrKrisKrisu:
> weil @NyCodeGHG und ich die einzigen beiden waren, die überhaupt das Flake benutzen (gehe ich von aus zumindest)

Do either of you actually use this? It seems to me like we have some unnecessary dead code lying around here that no one is actively maintaining. Can we remove the Nix support? Or is it still needed?

PR #3774, opened 2025-08-05
👽️ Use subdivision_code instead of region_code

### LINKED ISSUE #3758: Changes to the Transitous license.json format
### Describe the problem

We have started to replace the `region_code` and `region_name` properties with specific `country_` and `subdivision_` equivalents.
While the old field is still there for compatibility, we'd like to start removing it once Träwelling no longer relies on it.

PS.: You do not have a suitable issue template for these kind of requests.

### What steps did you take before the issue occurred?

   ​

### What did you expect to happen?

_No response_

### Browser console or error messages (optional)

```bash

```

### Any additional hints or information? (optional)

_No response_

HerrLevin:
Closes #3758

### DIFF
--- a/app/Console/Commands/FetchTransitousLicenses.php
+++ b/app/Console/Commands/FetchTransitousLicenses.php
@@ -38,7 +38,27 @@ class FetchTransitousLicenses extends Command
         $licenses = $this->getLicenseJson();
 
         foreach ($licenses as $license) {
-            $country    = $license['region_code'];
+            $country     = $license['country_code'] ?? null;
+            $subdivision = $license['subdivision_code'] ?? null;
+            if (empty($country)) {
+                $this->error('No country code found for license: ' . json_encode($license));
+                continue;
+            }
+
+            if (!empty($subdivision)) {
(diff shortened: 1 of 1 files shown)


app/Console/Commands/FetchTransitousLicenses.php
```diff
@@ -38,7 +38,27 @@ private function parseLicenses(): int {
...
-            $country    = $license['region_code'];
+            $country     = $license['country_code'] ?? null;
+            $subdivision = $license['subdivision_code'] ?? null;
+            if (empty($country)) {
+                $this->error('No country code found for license: ' . json_encode($license));
+                continue;
+            }
+
+            if (!empty($subdivision)) {
+                if (!str_starts_with($subdivision, $country . '-')) {
```
jbruechert, line 49: Is this a case that really happens? I think that would be a bug on the Transitous side.

HerrLevin: I don't really know. Just wanted to play it safe

PR #3802, opened 2025-08-17
:sparkles: add motis polylines

Quadrubo:
Closes #3705 

This uses https://github.com/Traewelling/GooglePolyline for converting the polyline data from motis to the traewelling format. I have tested this with the example from #3705 and another example without a provided polyline and both seem to work.

I had some issues with my code formatting ruining my diffs. Is there some tool you use to get your formatting style?

PR #3819, opened 2025-08-26
feat: export and upload geojson

Quadrubo:
Allows users to export the polyline as geojson and upload their own as a replacement.

Don't merge both this and #3802 as they change some of the same code. Will adjust one of them once one is merged.


MrKrisKrisu:
Hey, is this Draft-PR still up to date? I would love to see this features. :)

(but maybe check conflicts with #4024)

Quadrubo:
> Hey, is this Draft-PR still up to date? I would love to see this features. :)
> 
> (but maybe check conflicts with #4024)

Hey, i can rebase it to get it working on the weekend. In its current state it's not mergeable.

Quadrubo:
@MrKrisKrisu I remember why I didn't finish this now. Because of the way the polylines are shared between different users, when one user uploads a new one, with the current implementation, it changes it for all other users which is not fine. Don't really have time to approach this issue right now :/

Issue #1382, opened 2023-02-08
Checkins with a delay are sorted wrong


Oxbowatos:
**Describe the bug**
If e.g. I checkin in a train which derparting at 7am and I get on a train departing 6:30am but has 1 hour delay, the delayed checkin is sorted under the first checkin.

It also happens on the Global Dashboard and Home Dashboard that it's sorted wrong

**To Reproduce**

1. Checkin in a normal Train
2. At your Stop search for a delayed Train/Bus/etc (where the sheduled departure time is under the sheduled departure time of your first Checkin)
3. Look how the Checkins are sorted
4. See that the Checkin with Delay is sorted under the Checkins without Delay

**Expected behavior**
That the Page sorts the Checkins with the delayed departure time not only the sheduled departure time.

**Screenshots**

See how its sorted: (Deleted some fields for privacy reason)
![grafik](https://user-images.githubusercontent.com/123954576/217467007-4c456628-a154-4ec1-903b-9dfb7e0350a4.png)

**Desktop (please complete the following information):**
Any Browser

**Smartphone (please complete the following information):**
Any Smartphone

MrKrisKrisu:
This is our current query for the private dashboard:
https://github.com/Traewelling/traewelling/blob/74d3a988a149219d8371a1b0ff95d0ddc160d133/app/Http/Controllers/Backend/User/DashboardController.php#L19-L37

We sort here only according to the planned departure of the Journey:
https://github.com/Traewelling/traewelling/blob/74d3a988a149219d8371a1b0ff95d0ddc160d133/app/Http/Controllers/Backend/User/DashboardController.php#L27

In order for us to aggregate the data correctly directly we would need to do the following sorts in order:

1. if isset `DESC {join}.train_checkins.manual_departure`
2. `DESC {join}.train_checkins.{join}.train_stopovers.{real_departure}.`
2. `DESC {join}.train_checkins.{join}.train_stopovers.{planned_departure}.`


Since our data structure is a bit complex to build a meaningful and performant query for it I would suggest to do the sorting only on the frontend level so that it looks properly sorted to the user.

MrKrisKrisu:
ER-Diagram with the relevant tables:
![image](https://github.com/Traewelling/traewelling/assets/4103693/882747b7-cd27-4c45-b175-8357085a64c5)

PR #3832, opened 2025-08-29
✨Add Trusted User Status/tag visibility

sirlupusdev:
This is related to these discussions: #2028 #2959 

While the discussion as to whether private profiles should be able to post publicly is still ongoing (I assume), I've gone ahead and implemented this Trusted User visibility as proposed by @0AgentSmith in the discussions. This is shouldn't interfere with the "private profiles should not be able to post anything public" idea and it also doesn't require some sort of "migration" for current public check-ins of private profiles.

Tests have been added, I've tried to match them to the already existing ones (in terms of what to test/how to name things). Manual test with a dev instance was also done. 
Still, it'd be my first contribution to Träwelling, so I'm sorry if I forgot something somewhere.

### DIFF
--- a/app/Http/Controllers/Backend/Transport/StatusController.php
+++ b/app/Http/Controllers/Backend/Transport/StatusController.php
@@ -11,6 +11,7 @@ use App\Models\Stopover;
 use App\Models\User;
 use Closure;
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Facades\DB;
 
 abstract class StatusController extends Controller
 {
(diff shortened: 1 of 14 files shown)


app/Http/Controllers/Backend/Transport/StatusController.php
```diff
@@ -66,14 +67,30 @@ public static function filterStatusVisibility(?User $viewingUser = null): Closur
...
+                              $subQuery->select(\DB::raw(1))
+                                      ->from('trusted_users')
+                                      ->whereColumn('trusted_users.user_id', 'statuses.user_id')
+                                      ->where('trusted_users.trusted_id', $viewingUser->id)
+                                      ->where(function($expireQuery) {
+                                          $expireQuery->whereNull('trusted_users.expires_at')
+                                                     ->orWhere('trusted_users.expires_at', '>', now());
+                                      });
+                          });
+                });
```
MrKrisKrisu, line 93: I'm a little concerned that adding another `whereExists` to the query for the production data would make it even less performant. But... let's see.

(The query should generally be revised again, especially with regard to #1382 and #2703)

MrKrisKrisu (approved):
Thanks for the implementation! :)

My comment with the query is not blocking, so it should be looked at separately to see what can be optimized there...

MrKrisKrisu:
Dear @Traewelling/api-consumer, please check whether your implementation of the `StatusVisibility`-Enum is fail-safe; a new value has been added.

https://github.com/Traewelling/traewelling/blob/b992f02317dd8b70f24915437dc0f45a0112f5a3/app/Enum/StatusVisibility.php#L10-L11

app/Http/Controllers/Backend/Transport/StatusController.php
sirlupusdev: Yeah, I'm not sure how to handle it more efficiently otherwise (within the same query) though. One could obviously go and fetch all users that trust the current one (if logged in) and then filter it (again) after fetching the applicable rows, but I don't know whether that's doable or more efficient. 

Thank you for merging!

0AgentSmith:
I'd also like to express a big thank you for your implementation!
However, at least currently, it doesn't seem to work on my end. With experimental features on or off there is no option to display check-ins just for trusted users :/

sirlupusdev:
There's no release published for this commit yet. As soon as a release is published it should be available on the website as well. All releases can be found here: https://github.com/Traewelling/traewelling/releases

0AgentSmith:
Sorry, got a bit confused between _approved changes_ and _publishing a new release_ 
Thanks for clearing it up

PR #3869, opened 2025-09-08
:lock: reduce lifetime of access tokens

MrKrisKrisu:
see #2809

Suggested Changes:

- Access tokens now expire after 60 minutes
- Refresh tokens now expire after 30 days
- Personal access tokens now expire after 90 days

Notes for API developers:
- Existing tokens will keep their original expiration.
- Access Tokens can be refreshed every time within the period the refresh token is valid (30 days)


MrKrisKrisu:
Since there has been no further feedback here or in the discussion (#2809) for a few months, I assume that there are no more objections.

However, I have increased the lifetime of the access tokens to 60 minutes, which is a little more flexible.

@Traewelling/api-consumer

PR #3893, opened 2025-09-18
🐛 Trusted Users Fix: Alt 3, Add join instead of subquery for trusted users visibility

sirlupusdev:
This is a commit to be merged into the revert-revert branch, not develop directly (for now).

This uses joins to determine whether the checkin is to be viewed by the current user (basically by a row showing either the user id or null if not allowed). I'm not sure whether this would be an improvement of performance, but it'd most likely have to be used together with #3891. There's no reliable way for me to test based on the test data I have, although it may be faster than the approach using subqueries. The downside is that the join is done for every row regardless of the visibility of the checkin. 

Compatible with #3891, not #3892.


MrKrisKrisu:
performance on prod (with index from #3891) is okay-ish.

image

```csv
1	PRIMARY	s	range	PRIMARY,statuses_user_id_foreign,statuses_user_id_IDX,statuses_user_id_mastodon_post_id_created_at_index,statuses_user_id_visibility_index,idx_s_user_vis_created	statuses_user_id_visibility_index	9		348211	Using index condition; Using where; Using temporary; Using filesort
1	PRIMARY	tc	eq_ref	train_checkins_status_id_unique,train_checkins_departure_arrival_status_id_index,train_checkins_status_id_departure_index,idx_tc_departure_status	train_checkins_status_id_unique	8	traewelling.s.id	1	Using index condition; Using where
2	MATERIALIZED	tu	ref	trusted_users_user_id_trusted_id_unique,TEMP_FOR_PR_3857_TO_TEST	TEMP_FOR_PR_3857_TO_TEST	8	const	11	Using where; Using index
```

sirlupusdev:
With the join based query I assume? You could try comparing it to the Exists approach used before, the index is probably having the highest impact here. I think the subquery may only be executed when needed/the status is trusted whereas the join is done for all rows.

MrKrisKrisu:
> You could try comparing it to the Exists approach

it’s slightly faster and pulls in fewer rows as the join query. I'll push it in a few minutes. I'm not sure what causes the DB-struggles.

image

```txt
1	PRIMARY	statuses	range	PRIMARY,statuses_user_id_foreign,statuses_user_id_IDX,statuses_user_id_mastodon_post_id_created_at_index,statuses_user_id_visibility_index,idx_s_user_vis_created	statuses_user_id_visibility_index	9		200303	Using index condition; Using where; Using temporary; Using filesort
1	PRIMARY	train_checkins	eq_ref	train_checkins_status_id_unique,train_checkins_departure_arrival_status_id_index,train_checkins_status_id_departure_index,idx_tc_departure_status	train_checkins_status_id_unique	8	traewelling.statuses.id	1	Using index condition; Using where
2	MATERIALIZED	trusted_users	ref	trusted_users_user_id_trusted_id_unique,TEMP_FOR_PR_3857_TO_TEST	TEMP_FOR_PR_3857_TO_TEST	8	const	11	Using where; Using index
```

MrKrisKrisu (approved):

sirlupusdev:
The exists query is already implemented in the branch (as part of my initial PR and that revert), so only the alt 1 PR would be needed as far as I'm aware.

PR #4024, opened 2025-10-24
✨ replace Brouter by OpenRailRouting and split polylines into segments

HerrLevin:
This PR aims to fix the current polyline situation by retreiving data from https://routing.openrailrouting.org/, splitting travel distances between multiple stops and stitching it together in the statuses.

- [x] Fetching data from openrailrouting
- [x] Map-Matching request, to reduce false-positives in route-snapping
- [x] Calculating distance based on route segments
- [x] Recalculating all changed posts
- [ ] ~~Only fetch stations, if more than x percent of stopovers don't have a route segment~~


app/Services/OpenRailRoutingService.php
```diff
@@ -0,0 +1,116 @@
...
+
+        $url      = sprintf(
+            '%s/match?type=json&key=&elevation=false&instructions=false&profile=%s',
+            self::API_URL,
+            $profile->value,
+        );
+        $response = $this->getClient()->post($url, [
+            'headers' => ['Content-Type' => 'application/gpx+xml'],
+            'body' => $gpxFile->toXML()->saveXML()
+        ]);
```
MrKrisKrisu, line 68: Please add an timeout here.

Image

app/Jobs/RefreshPolyline.php
```diff
@@ -2,26 +2,49 @@
...
+use Cache;
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Bus\Dispatchable;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Facades\Log;
 use romanzipp\QueueMonitor\Traits\IsMonitored;
 
 class RefreshPolyline implements ShouldQueue
```
MrKrisKrisu, line 16: Please skip this task if a polyline is specified in the schedule, as otherwise we will destroy the correct data again.

MrKrisKrisu: Ah, nvm. Seems to be checked here:
https://github.com/Traewelling/traewelling/pull/4024/files#diff-f9105698aa1ab40e6297377ba2e807d79f77ec189c42dd1510f209d7e2ad950eR45

PR #4029, opened 2025-10-25
👽️ Use Motis API v5 wherever possible

HerrLevin:


kasmar00:
Hi, api v4 introduced `displayName` property, which contains what `routeShortName` had in api v1 ([source](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/motis-project/motis/refs/heads/master/openapi.yaml#:~:text=/api/v4/%7Bplan%2Ctrip%2Cstoptimes%2Cmap/trips%7D%20new%20displayName%20property%2C%20routeShortName%20only%20contains%20actual%20route%20short%20name%20from%20source%3B%20since%20MOTIS%202.2.0)).
After this PR (for example) Polish trains are displayed as `IC` instead of `IC 1234` (as was before). Example: old: https://traewelling.de/status/6174619, new: https://traewelling.de/status/6186992

Is this intended change? If not, then probably traewelling should switch to use `displayName` instead of `routeShortName`.
(If you'd preffer I open this as a sepperate issue let me know, if you agree with changing do `displayName` I can try to do it myself and open a PR).

HerrLevin:
@kasmar00 Thank you so much! I completely missed that in the documentation. I'll fix it switfly.

PR #4096, opened 2025-11-04
:art: refactor station repository and controller to use station identifiers; remove deprecated methods

MrKrisKrisu:


MrKrisKrisu:
Dear @Traewelling/api-consumer,

A (currently non-breaking) change:

We are currently migrating the existing identifiers (Wikidata ID, IBNR, Ril100) to a separate model in order to standardize everything.

With this PR, we are already outputting the stationIdentifier in the new format in many API endpoints.

The old fields still exist, but will return `null` in the medium term until they are removed in the long term.

So please feel free to slowly convert your applications (after the release) to the new attribute if you use some of the identifiers.

Currently we have these four types of identifier (but please don't strictly use this enum, as we add more and more here...):

https://github.com/Traewelling/traewelling/blob/2425b99543718ac16c5f765788cbc19ea7434003/app/StationIdentifierType.php#L7-L10

For the autocomplete endpoint for example:

image

As of the updated API documentation we'll remove the columns to this format:

image

marhei:
@MrKrisKrisu, there are now two distinct Station resources mentioned in the API documentation. 

Are similar chances planned for the stopover resource?

PR #4127, opened 2025-11-11
:sparkles: add CSV import for trip creation form

MrKrisKrisu:


wittighausen:
Is it planned to expand this feature, so the operators and line-/train number could be imported this way too among with the stations?
Especially because the amount of the available and repeating operators can make some manually created Check-Ins not consistent in the meaning of operators (alone DB Fernverkehr AG is available 11 times, so it would be better if you could stick to one of them then guess which one was the previous used and e.g. have 4 "different" operators in your statistics, which are actually the same)

MrKrisKrisu:
> Is it planned to expand this feature

I don't plan to implement this. But if anyone wants to, feel free. :)

I've grouped the DB Fernverkehr operators:
image

PR #4256, opened 2025-12-25
✨ add customizable `route_text_color` string

wittighausen:
Basically changes the code, so the custom color of the line text can be shown too.
Mainly done because of problems with lines, which only have a text color and are saved with a background color `#ffffff`, which now result in an automatic `#000000` as the text color.

Just as an example of what this is supposed to do:
How the line icon looks right now
Screenshot 2025-12-25 at 02-21-46 whygotobayern&#39;s Reise mit Träwelling - Träwelling

How the line icon should look afterwards
Screenshot 2025-12-25 at 02-21-46 whygotobayern&#39;s Reise mit Träwelling - Träwelling

(Just to be clear: I haven't done this type of thing in my entire life, so I just hope this all works and I'll be happy)


MrKrisKrisu (commented):
Thanks for the changes. It looks good overall. However, the colors are not saved because, for example, the column is still missing in the database (I don't see a migration here). This means that the colors are only displayed in the check-in flow and not in the existing status.

resources/vue/components/Checkin/StationBoardEntry.vue
```diff
@@ -73,7 +70,7 @@ const backgroundColor = computed<string | null>(() => {
 const textColor = computed<string | null>(() => {
   // Nur setzen, wenn tatsächlich ein valider Background vorliegt
   if (!backgroundColor.value) return null;
-  const c = contrastTextColor(rawApiHex.value);
+  const c = routeTextColorCss(rawApiHex.value);
```
MrKrisKrisu, line 73: But something is wrong here. The background color is the same as the text color.

Image

Image

MrKrisKrisu: API response is correct: 
image

wittighausen:
So I've adjusted some things and hopefully it'll work now as it should

tests/Feature/APIv1/CheckinTest.php
```diff
@@ -18,7 +18,8 @@ class CheckinTest extends ApiTestCase
     public function testOauthClientIdIsSavedOnApiCheckins(): void {
         $user  = User::factory()->create();
         $token = $user->createToken('token', array_keys(AuthServiceProvider::$scopes));
-        $trip  = Trip::factory()->create();
+        $client = \App\Models\Client::factory()->create(['id' => 1]);
+        $trip   = Trip::factory()->create(['client_id' => $client->id]);
```
MrKrisKrisu: This test doesn't belong to your changes, or? Can you please revert this?

resources/vue/components/Checkin/StationBoardEntry.vue
MrKrisKrisu, line 73: Thanks, now it's working and the correct color is rendered. I'm not fully happy with the dataset, but this is a maybe a thing for line-colors and/or transitous.

Image

MrKrisKrisu (changes requested):

MrKrisKrisu (approved):

resources/vue/components/Checkin/StationBoardEntry.vue
wittighausen: @MrKrisKrisu just as a quick question, should the line icons look like this? 
For me, the colors seem a bit off, even though they are saved correctly in the data.
Screenshot 2026-01-12 at 18-08-16 RIS - Träwelling

MrKrisKrisu: continue conversation at #4291

PR #4357, opened 2026-01-24
:recycle: migrate identifiers; remove legacy

MrKrisKrisu:


MrKrisKrisu:
I think the PR is a bit big in one go. I'll break it down into smaller steps soon.

PR #4513, opened 2026-03-05
:sparkles: add experimental route editor in backend

MrKrisKrisu:
(currently for testing purposes. should be in contribution system later)


MrKrisKrisu:
Since OpenRailRouting has unfortunately caused more problems than it has solved so far, I have decided to test BRouter again (as we used to do in the past).


## With OpenRailRouting

grafik

## With BRouter

grafik

PR #4624, opened 2026-03-25
♻️ Refactor changelog logic

HerrLevin:


app/Http/Controllers/Backend/ChangelogController.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace App\Http\Controllers\Backend;
```
MrKrisKrisu, line 3: Maybe move to a service?

HerrLevin: I'll think about it once I'm implementing the api endpoint

PR #4692, opened 2026-04-12
:bug: explicitly set scopes for new mastodon instances

### LINKED ISSUE #2666: Too many requested scopes
### Describe the bug

Träwelling is requesting too many scopes on activitpub-servers ~that are not mastodon.~

Due to a recent discussion in #2768 and subsequent debugging I've come to the following conclusion:
> The problem is the following: Our old mastodon library had a bug and didn't allow anything except read/write scopes. It's technically possible now to change it, but we would have to rewrite a huge chunk of our login/social posting code to handle both "old" keys with their scopes and the new ones with reduced scopes. Also we would have to re-authenticate all users and about 500 instances. Which is a lot.

> [!NOTE]
> What should we do here? I'm not sure if we can and should still change this.
> Maybe add a flag in the mastodon servers table indicating the scopes?

### Steps to reproduce

1. Create an account on a misskey instance
2. Log in on TRWL with your account
3. See way more scopes than read/write

### Browser console logs

```bash
See also: https://blahaj.zone/notes/9u5sdaa31l6o01la
```

### Browser

All

MrKrisKrisu:
fixes #2666

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -51,6 +51,7 @@ MASTODON_ID=
 MASTODON_SECRET=
 MASTODON_REDIRECT=
 MASTODON_APPNAME=
+MASTODON_SCOPES="read:statuses write:statuses read:accounts"
 
 # Default values to be shown in the legal notice.
 MIX_LEGAL_NAME="Max Mustermann"
--- a/app/Http/Controllers/Backend/Social/MastodonController.php
+++ b/app/Http/Controllers/Backend/Social/MastodonController.php
@@ -75,9 +75,9 @@ abstract class MastodonController extends Controller
     {
         try {
             $info = Mastodon::domain($domain)->createApp(
-                client_name: config('services.mastodon.client_name'), // TODO: why is client name required here?
+                client_name: config('services.mastodon.client_name'),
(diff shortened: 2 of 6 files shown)


HerrLevin:
Code wise this looks good to me.

But have you tested this locally? I mean, did you create an app first and then log in again after making your changes—or try it with a different one?

If not, please try it with multiple instances. Mastodon, Pleroma, GoToSocial, MissKey, etc.
OAuth2 can be tricky to implement. Sometimes changing a scope can break things.

MrKrisKrisu:
> If not, please try it with multiple instances. Mastodon, Pleroma, GoToSocial, MissKey

I can test that further. But since when have we supported anything other than Mastodon? We’ve never claimed to support anything else.

HerrLevin:
> But since when have we supported anything other than Mastodon? We’ve never claimed to support anything else.

I know that, but we still have a huge userbase with very specific activitypub configurations (and many without an email address). I'd rather not see them locked out of their account + us having to fix it manually. :D

HerrLevin:
LGTM

PR #4695, opened 2026-04-12
:card_file_box: delete duplicate checkins; add unique key

### LINKED ISSUE #2876: Duplicate checkins possible
### Describe the bug

As a user I currently can checkin multiple times to the same journey. People using the friend checkin causes duplicate checkins multiple times.

(I thought this is fixed with a unique db constraint, but need to be checked.)

### Steps to reproduce

1. Check into a journey.
2. Check in again.

### Browser console logs

_No response_

### Browser

_No response_

MrKrisKrisu:
fixes #2876

I would like to add a UNIQUE constraint to the database to prevent duplicate check-ins. Unfortunately, this has been possible (again) since 2024, as the UNIQUE constraint was accidentally deleted.

My approach:

This migration checks for duplicate check-ins. (= A user checked in at the same station at the same time during a trip.)

The oldest check-in (lowest ID) is keept; the others are deleted.

However, I don’t want to run this without first opening it up for discussion. Do you see any issues with this?


MrKrisKrisu:
Note: This PR is currently on hold. To avoid deleting user data without permission, I would like to create a separate PR to give users with duplicate check-ins the opportunity to clean this up themselves first. I would then notify the users and set a deadline. After the deadline, I would merge this PR and delete the duplicate check-ins.

MrKrisKrisu:
Blog post published: https://blog.traewelling.de/posts/duplicate-checkins/

PR #4806, opened 2026-05-18
:sparkles: add alert for upcoming privacy change

MrKrisKrisu:


MrKrisKrisu:
Hello @Traewelling/api-consumer,

Just a quick heads-up: We will be update our Privacy Policy in the coming days/weeks/months (...) - see https://github.com/Traewelling/traewelling/pull/4764.

As a result, all users will need to confirm their acceptance of the new policy. There will be a transition period between the announcement and the point at which confirmation becomes mandatory.

If you have an application with many users, we recommend that you integrate the API endpoint for alerts so that your users aren’t suddenly surprised when the API returns an HTTP 403 error.

Example Response of Endpoint `/api/v1/alerts`: 

```json
{
  "data": [
    {
      "id": "privacy-policy-upcoming",
      "type": "warning",
      "active_from": "2026-05-18T17:23:27.000000Z",
      "active_until": "2026-12-31T00:00:00.000000Z",
      "url": null,
      "translations": [
        {
          "title": "New Privacy Policy",
          "content": "A new privacy policy will take effect on December 31, 2026. You can accept it early.",
          "url": "http://localhost:8000/gdpr-intercept",
          "locale": "en"
        }
      ]
    }
  ]
}
```

Also, please note: We may also use the Alert Endpoint to display information about scheduled maintenance. Or, if our Year in Review feature is active, that information will also be displayed there.

The available `type` values are `info`, `warning`, `danger` and `success`. It would be sufficient for you to accept `warning` and`danger`. `info` and `success` are generally irrelevant for third-party apps.

PR #4891, opened 2026-06-18
🐛 Fix HTTP 500 error when sending webhooks that have intentionally missing fields

DRSchlaubi:
Currently there is an HTTP 500 when e.g. editing statuses which have an event set, this bug only affects only users with at least one `checkin_update` webhook.

This is caused by the newly introduced `freezeJsonObjects` function, which tries to call `toArray` on those resources, which fails, because for instances of `MissingValue` `JsonResource->$resource` is null. Looks like my security fix broke this

In my defense, if it would not be `freezeJsonObjects` it would be the DefaultSigner throwing the same exception

I still have to validate whether signature still pass after this

#HSFDPMUW


DRSchlaubi:
Signature is generated as expected

MrKrisKrisu (approved):

PR #4954, opened 2026-07-18
:bug: don't map real time data with identifiers

### LINKED ISSUE #4948: Duplicated Stopovers
### Are you sure this is not a data issue?

- [x] Yes, this is a Traewelling software bug (not a data issue).

### Have you checked for existing issues?

- [x] Yes, I have verified that no existing issue covers this problem.

### Affected URL (if any)

_No response_

### Describe the problem

For some trips stopovers seem to be duplicated, seems to happen between Basel and Mannheimt a lot, but it also occurred on the S6 of the S-Bahn munich for me

### Steps to reproduce

https://traewelling.de/api/v1/stopovers/8167523
https://api.transitous.org/?tripId=20260715_14%3A24_de-DELFI_3288192754

Stopover planegg is duplicated

### Expected behavior

_No response_

### Environment

_No response_

### Browser console / error messages (optional)

```bash

```

### Additional info (optional)

_No response_

MrKrisKrisu:
fixes #4948

### DIFF
--- a/app/DataProviders/Repositories/TripRepository.php
+++ b/app/DataProviders/Repositories/TripRepository.php
@@ -21,11 +21,11 @@ class TripRepository
                 $trip->stopovers()->updateOrCreate(
                     [
                         'trip_id' => $trip->trip_id,
-                        'train_station_id' => $stopover->station->id,
                         'arrival_planned' => $stopover->arrival_planned,
                         'departure_planned' => $stopover->departure_planned,
                     ],
                     [
+                        'train_station_id' => $stopover->station->id,
                         'arrival_real' => $stopover->arrival_real,
                         'departure_real' => $stopover->departure_real,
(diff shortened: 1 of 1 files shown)


MrKrisKrisu:
> I'm not entirely sure what we can do here (other than improving our data quality, but that's an endless task and involves a lot of manual work).
> 
> Maybe we could remove the StationID from the matching process and, when updating, only match based on trip_id and planned_times.
>
> Right now, I can’t think of any edge cases that would break as a result. I’ve already finished the PR for this, but I won’t merge it just yet in case anyone has any vetoes.
>
> _originally posted by @MrKrisKrisu in #4948_

MrKrisKrisu:
I guess we should test it. I've seen many matching problems today as well that can be fixed with this PR. If anyone see any problems, please report.

PR #5023, opened 2026-08-16
✨ add stationboard timestamp in status

pa-ulex:
This PR adds a button to the status context menu, making it quicker and easier to add connections in advance.


pa-ulex:
I unfortunatly can not reach Weblate...

MrKrisKrisu:
Thanks for the PR! Would it be an alternatative option to add the timestamp to search to the link when clicking the origin or destination in a status (=timestamp of arrival / departure)?

I would prefer to see as less action elements in the context menu as possible.

pa-ulex:
Sure, that's a brilliant solution!

pa-ulex:
I have only added the timestamp for the next stop and the destination. I don't think the timestamp is needed for departure, as it's easy to change them.

MrKrisKrisu:
I've added the timestamp also to the origin. It would be confusing to have two links where the logic behind is not the same.

I also added this change to our new frontend. You just added it to our legacy frontend which will be removed soon.

MrKrisKrisu (approved):

west-train:
Is it possible to make it dependent on whether the journey has already arrived? Imagine a journey arriving at 10:00, so if you open the destination stationboard at 9:30, it will skip directly to 10:00, but if you open it at 10:30, it will remain at the current time? I definitely see the advantage of this new feature, but in case of very long interchange times (as I just experienced today), it actually needs two more clicks than before to find my connection.

d3d9:
I agree, while this seems useful for some cases, in others I find myself having to change the time because I actually want the current departures. The "Now" button next to the picker definitely helps though. I know that I can also query for a previous stop by clicking on the search bar, but that can take some time to load/appear especially on a spotty connection while my previous status is often just there on the previously loaded page.
The time-based logic could also lead to unexpected effects depending on where the threshold would be, so how about making the times in the status clickable, clicking on those would open the stationboard with that time set while clicking on the stop name wouldn't? Could still be considered too obscure or something but I wanted to try to suggest this.

PR #5030, opened 2026-08-21
:bug: remove 5 minute offset in stationboard requests

MrKrisKrisu:


d3d9:
Oh, this behavior was a 🐛? I catch myself having to manually change the queried time by a few mins a lot since this change 😳
The way past departures are grayed out also helps quickly identify the departures for the queried time for those who actually don't care about the slightly earlier ones, so maybe this could be reverted or changed in another way?

edit: This also particularly applies in combination with the status timestamp from #5023, sometimes you manage to catch a connection e.g. due to its slight delay or your slightly earlier arrival (which might not be reflected in the data in time due to usual latency/accuracy things), the previous offset really helped in these cases

PR #5050, opened 2026-08-29
:recycle: document the API version per endpoint instead of in the server url

MrKrisKrisu:


docs/architecture/decisions/0002-per-endpoint-api-versioning.md
```diff
@@ -0,0 +1,67 @@
...
+
+- A client author should be able to see what an endpoint returns without cross-referencing dates
+- One endpoint, one answer: no field that means different things before and after a deadline
+- Existing clients must keep working until they have had a fair chance to migrate
+- Migration should be a single, discrete step a client can plan, not a slow drift
+
+## Decision Outcome
+
+**When an endpoint has to break, it gets a new version. The old version stays as it is and is
+switched off later, with notice.**
```
HerrLevin, line 34: idk if this is useful in this adr or if it should be added to another pr:
We should introduce a so called "sunset header" to also display the EOL of one endpoint: https://www.rfc-editor.org/info/rfc8594/

MrKrisKrisu: I like the idea (TIL).

